PR #6, opened 2023-06-17
Add generic building graphics fallback

### LINKED ISSUE #5: Lacks NewTheater support for generic (xGxxxx) artwork, to go along with the defined NewTheaterBuildingLetter
I don't know the solution here, but the fallback to "generic" doesn't seem to be supported. Snow maps load all the player buildings, but urban maps won't, when NewTheaterBuildingLetter=U.

Either make NewTheaterBuildingLetter into a list, like `NewTheaterBuildingLetter=U,G` to indicate priority, or add a separate tag like `EnableGenericFallback=yes` to accomplish the same.

MortonPL:
When searching for building shapes, RA2 and YR can fall back to generic letter "G" when no theater-specific file can be found. I.e. when no `GAPILE.shp` can be found, `GGPILE.shp` will be used instead. This PR allows users to enable this behavior. Closes #5.

In `Config/Constants.ini`:
```ini
[Constants]
NewTheaterGenericBuilding=false; boolean
```

### DIFF
--- a/src/TSMapEditor/Constants.cs
+++ b/src/TSMapEditor/Constants.cs
@@ -14,6 +14,7 @@ namespace TSMapEditor
 
         public static bool IsFlatWorld = false;
         public static bool TheaterPaletteForTiberium = false;
+        public static bool NewTheaterGenericBuilding = false;
 
         public static string ExpectedClientExecutableName = "DTA.exe";
         public static string GameRegistryInstallPath = "SOFTWARE\\DawnOfTheTiberiumAge";
@@ -72,6 +73,8 @@ namespace TSMapEditor
         public const string ClipboardMapDataFormatValue = "ScenarioEditorCopiedMapData";
         public const string UserDataFolder = "UserData";
 
+        public const char NewTheaterGenericLetter = 'G';
+
         public static void Init()
         {
             const string ConstantsSectionName = "Constants";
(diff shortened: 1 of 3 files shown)


src/TSMapEditor/Constants.cs
```diff
@@ -92,6 +95,7 @@ public static void Init()
             TheaterPaletteForTiberium = constantsIni.GetBooleanValue(ConstantsSectionName, nameof(TheaterPaletteForTiberium), TheaterPaletteForTiberium);
             ExpectedClientExecutableName = constantsIni.GetStringValue(ConstantsSectionName, nameof(ExpectedClientExecutableName), ExpectedClientExecutableName);
             GameRegistryInstallPath = constantsIni.GetStringValue(ConstantsSectionName, nameof(GameRegistryInstallPath), GameRegistryInstallPath);
+            NewTheaterGenericBuilding = constantsIni.GetBooleanValue(ConstantsSectionName, nameof(NewTheaterGenericBuilding), NewTheaterGenericBuilding);
```
Rampastring: Move this to line 96 so the fields are ordered similarly to their definitions on lines 15 to 20.

Rampastring (changes requested):
Looks good aside from one nitpick.

Rampastring (approved):

PR #7, opened 2023-06-17
Add Ares/Phobos style INI include for game INIs

MortonPL:
This PR adds support for [Ares](https://ares-developers.github.io/Ares-docs/new/misc/include.html) and [Phobos](https://phobos.readthedocs.io/en/latest/Miscellanous.html#include-files) INI include feature and [Phobos INI section inheritance](https://phobos.readthedocs.io/en/latest/Miscellanous.html#section-inheritance) feature for rules, art, theme and theater INIs. Map editor config files are not affected by this change.

Wasn't sure where to put this new `IniFileEx`, so I created a temporary directory `Extensions`.

Example use case of linked features:

`rulesmd.ini`:
```ini
[$Include]
0=new_rules.ini
```

`new_rules.ini`:
```ini
[E2]
$Inherits=SOVIET_OWNERS
; no Owner defined here

[SOVIET_OWNERS]
Owner=Russians,Confederation,Africans,Arabs
```

### DIFF
--- a/src/TSMapEditor/Extensions/IniFileEx.cs
+++ b/src/TSMapEditor/Extensions/IniFileEx.cs
@@ -1,0 +1,30 @@
+﻿using System.IO;
+using System.Text;
+using Rampastring.Tools;
+
+namespace TSMapEditor.Extensions;
+
+public class IniFileEx: Rampastring.Tools.IniFile
+{
+    public IniFileEx() : base() { }
+
+    public IniFileEx(string filePath) : base(filePath) { }
+
+    public IniFileEx(string filePath, Encoding encoding) : base(filePath, encoding) { }
+
+    public IniFileEx(Stream stream) : base(stream) { }
+
+    public IniFileEx(Stream stream, Encoding encoding) : base(stream, encoding) { }
+
+    public new void Parse()
+    {
+        base.Parse();
+
+        foreach (var pair in GetSection("$Include").Keys)
+        {
+            string path = SafePath.CombineFilePath(SafePath.GetFileDirectoryName(FileName), pair.Value);
+            IniFileEx includedIni = new(path);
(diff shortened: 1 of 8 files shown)


src/TSMapEditor/Extensions/IniFileEx.cs
```diff
@@ -0,0 +1,86 @@
...
+            if (!section.KeyExists(PhobosInheritsSection))
+                continue;
+
+            foreach (var parentName in section.GetListValue<string>(PhobosInheritsSection, ',', x => x))
+            {
+                var parent = Sections.Find(s => s.SectionName == parentName);
+                if (parent == null)
+                    continue;
+
+                foreach (var pair in parent.Keys)
```
Rampastring, line 86: Add braces for this `foreach`. Don't nest braceless expressions.

Rampastring (changes requested):
One stylistical issue, looks good otherwise.

I see that it only supports Ares OR Phobos include sections, you can't use both at the same time in YR?

I also wonder whether going through all sections like this can have a performance hit on map (incl. rules) load time. If yes, this should be made an optional toggle. Although the effect is probably insignificant enough to not really matter. Have you observed any difference?

Metadorius:
> Ares OR Phobos include sections, you can't use both at the same time in YR?

Correct. The hooks conflict with each other and enabling Phobos implementation disables Ares one.

MortonPL:
> I also wonder whether going through all sections like this can have a performance hit on map (incl. rules) load time. If yes, this should be made an optional toggle. Although the effect is probably insignificant enough to not really matter. Have you observed any difference?

It wasn't noticeable for vanilla YR rules, but I've made this addition opt-in anyway.

Rampastring (approved):

PR #12, opened 2023-06-27
Add view drag scrolling

MortonPL:
This PR adds an alternate way to move map view around. Right click scroll is disabled when using drag scroll and vice versa. Zooming is disabled when using drag scroll.

This action is bound to middle mouse button (and because of that, this PR depends on [this XNAUI PR introducing middle mouse button input handling](https://github.com/Rampastring/Rampastring.XNAUI/pull/28)).


Rampastring:
This needs a rebase.

PR #13, opened 2023-06-29
Encrypted MIXes support

MortonPL:
This PR adds support for encrypted MIXes by implementing RSA and Blowfish decryption. No external dependencies are introduced.

MIXes with checksum are accepted too, since they can be read as usual if you ignore the checksum at the end. Implementing checksum validation is currently impossible, because it requires linear reading of the entire MIX body, while this project uses random access reading.

### DIFF
--- a/src/TSMapEditor/CCEngine/MixFile.cs
+++ b/src/TSMapEditor/CCEngine/MixFile.cs
@@ -1,6 +1,8 @@
-﻿using System;
+﻿using SharpDX.Direct3D11;
+using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 
 namespace TSMapEditor.CCEngine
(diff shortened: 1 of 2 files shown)


src/TSMapEditor/CCEngine/MixFile.cs
```diff
@@ -85,8 +86,77 @@ public void Parse(Stream stream = null)
...
+                ParseUnencrypted(stream);
+
+            if (masterMix != null)
+                masterMix.CloseFile();
+        }
+
+        /// <summary>
+        /// Reads MIX file entry information from a Blowfish encrypted stream.
+        /// </summary>
+        /// <param name="stream"></param>
```
Rampastring: Rather leave out the comment for this parameter (it's self-explanatory) than leave it empty.

src/TSMapEditor/CCEngine/MixFile.cs
```diff
@@ -85,8 +86,77 @@ public void Parse(Stream stream = null)
...
+                entries.Add(new MixFileEntry(buffer, dataOffset));
+
+                dataOffset += MixFileEntry.SIZE_OF_FILE_ENTRY;
+            }
+        }
+
+        /// <summary>
+        /// Reads MIX file entry information from an unencrypted stream.
+        /// </summary>
+        /// <param name="stream"></param>
```
Rampastring: Rather leave out the comment for this parameter (it's self-explanatory) than leave it empty.

Rampastring (changes requested):
Looks good, a couple of small comments.

MortonPL:
Moved decrypted data buffering inside BlowfishStream and reworked the code a bit so that `Parse()` logic is almost the same with and without encryption.

src/TSMapEditor/CCEngine/MixFile.cs
```diff
@@ -84,16 +84,29 @@ public void Parse(Stream stream = null)
             byte[] buffer = new byte[256];
             stream.Read(buffer, 0, 4);
             MixType mixType = (MixType)BitConverter.ToInt32(buffer, 0);
+            
+            var isEncrypted = (mixType & MixType.ENCRYPTED) != 0;
```
Rampastring: Don't use `var` for primitive types.

Rampastring (changes requested):

Rampastring (approved):

PR #14, opened 2023-06-29
Custom tiberium/ore palette per theater

MortonPL:
This PR allows users to set a custom tiberium/ore palette for each theater with `TiberiumPaletteName`. This setting is respected only when `[Constants]▶TheaterPaletteForTiberium=true`. If `TiberiumPaletteName` is not set, theater palette will be used.

RA2/YR uses a separate palette for ores (and possibly other things that I am unaware of), i.e. for temperate it's `isotem.pal` and `temperat.pal`.

In `Theater.ini`:
```ini
[SOMETHEATER]
TiberiumPaletteName=      ; palette file name with extension, defaults to nothing
```

### DIFF
--- a/src/TSMapEditor/Config/Theaters.ini
+++ b/src/TSMapEditor/Config/Theaters.ini
@@ -13,6 +13,7 @@ ConfigINIPath=INI/Temperat.ini
 ContentMIXName=IsoTemp.mix
 TerrainPaletteName=isotem.pal
 UnitPaletteName=unittem.pal
+TiberiumPaletteName=
 FileExtension=.tem
 NewTheaterBuildingLetter=A
 
--- a/src/TSMapEditor/Config/Theaters.ini
+++ b/src/TSMapEditor/Config/Theaters.ini
@@ -13,7 +13,7 @@ ConfigINIPath=INI/Temperat.ini
 ContentMIXName=IsoTemp.mix
 TerrainPaletteName=isotem.pal
 UnitPaletteName=unittem.pal
-TiberiumPaletteName=
+;TiberiumPaletteName=
 FileExtension=.tem
 NewTheaterBuildingLetter=A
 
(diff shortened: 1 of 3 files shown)


src/TSMapEditor/Rendering/TheaterGraphics.cs
```diff
@@ -422,6 +422,8 @@ public TheaterGraphics(GraphicsDevice graphicsDevice, Theater theater, CCFileMan
 
             theaterPalette = GetPaletteOrFail(theater.TerrainPaletteName);
             unitPalette = GetPaletteOrFail(Theater.UnitPaletteName);
+            if (Theater.TiberiumPaletteName.Length > 0)
```
Rampastring: Use `string.IsNullOrWhiteSpace`. That, as its name says, also covers cases when the string value is null or white space rather than just "empty". This effectively also makes the INI key optional.

src/TSMapEditor/Config/Theaters.ini
```diff
@@ -13,6 +13,7 @@ ConfigINIPath=INI/Temperat.ini
 ContentMIXName=IsoTemp.mix
 TerrainPaletteName=isotem.pal
 UnitPaletteName=unittem.pal
+TiberiumPaletteName=
```
Rampastring: Please comment this out as it's not necessary for DTA. By leaving it commented out, it still acts as a reference for YR modders.

src/TSMapEditor/Rendering/TheaterGraphics.cs
```diff
@@ -813,7 +815,12 @@ public void ReadOverlayTextures(List<OverlayType> overlayTypes)
                     shpFile.ParseFromBuffer(shpData);
                     Palette palette = theaterPalette;
 
-                    if (overlayType.Wall || overlayType.IsVeins || (overlayType.Tiberium && !Constants.TheaterPaletteForTiberium))
+                    if (overlayType.Tiberium && Constants.TheaterPaletteForTiberium)
+                        palette = tiberiumPalette ?? theaterPalette;
+                    else if (overlayType.Tiberium)
+                        palette = unitPalette;
```
Rampastring, line 820:
```cs
if (overlayType.Tiberium)
{
    palette = unitPalette;
    
    if (Constants.TheaterPaletteForTiberium)
        palette = tiberiumPalette ?? theaterPalette;
}
```

This avoids the duplicate check for `overlayType.Tiberium`.

Rampastring (changes requested):

Rampastring (approved):

PR #15, opened 2023-06-30
Allow reading game INIs from MIX files

MortonPL:
This PR enables reading rules, expansion rules, art, expansion art, theme and theater INI files directly from MIXes. Files in game directory remain higher priority.
Tutorial INI is not included, as it uses a file watcher.

This PR also makes building foundation parsing less strict (case insensitive).

### DIFF
--- a/src/TSMapEditor/Extensions/IniFileEx.cs
+++ b/src/TSMapEditor/Extensions/IniFileEx.cs
@@ -49,6 +49,9 @@ public class IniFileEx: IniFile
     /// <returns>Constructed INI file or empty INI file on failure.</returns>
     public static IniFileEx FromPathOrMix(string filePath, string gameDirectory, CCFileManager ccFileManager)
     {
+        if (filePath.Length == 0)
+            return new();
+
         string rulesPath = Path.Combine(gameDirectory, filePath);
         if (File.Exists(rulesPath))
             return new(rulesPath);
--- a/src/TSMapEditor/Extensions/IniFileEx.cs
+++ b/src/TSMapEditor/Extensions/IniFileEx.cs
@@ -48,19 +48,19 @@ public class IniFileEx: IniFile
     /// <param name="filePath">Path to file or file name inside MIX file system.</param>
     /// <param name="gameDirectory">The path to the game directory.</param>
     /// <param name="ccFileManager">File manager object holding MIXes.</param>
-    /// <returns>Constructed INI file or empty INI file on failure.</returns>
+    /// <returns>Loaded INI file, or empty INI file object if the file was not found.</returns>
     public static IniFileEx FromPathOrMix(string filePath, string gameDirectory, CCFileManager ccFileManager)
     {
         if (filePath.Length == 0)
             return new();
 
-        string rulesPath = Path.Combine(gameDirectory, filePath);
-        if (File.Exists(rulesPath))
-            return new(rulesPath, ccFileManager);
+        string iniPath = Path.Combine(gameDirectory, filePath);
+        if (File.Exists(iniPath))
+            return new(iniPath, ccFileManager);
 
-        var rulesBytes = ccFileManager.LoadFile(filePath);
-        if (rulesBytes != null)
-            return new(new MemoryStream(rulesBytes), ccFileManager);
+        var iniBytes = ccFileManager.LoadFile(filePath);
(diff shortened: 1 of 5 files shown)


src/TSMapEditor/Extensions/IniFileEx.cs
```diff
@@ -15,34 +16,59 @@ public class IniFileEx: IniFile
...
+    /// <param name="filePath">Path to file or file name inside MIX file system.</param>
+    /// <param name="gameDirectory">The path to the game directory.</param>
+    /// <param name="ccFileManager">File manager object holding MIXes.</param>
+    /// <returns>Constructed INI file or empty INI file on failure.</returns>
+    public static IniFileEx FromPathOrMix(string filePath, string gameDirectory, CCFileManager ccFileManager)
+    {
+        if (filePath.Length == 0)
+            return new();
+
+        string rulesPath = Path.Combine(gameDirectory, filePath);
```
Rampastring: `rulesPath` is not a good name as this might be any INI file.

src/TSMapEditor/Extensions/IniFileEx.cs
```diff
@@ -15,34 +16,59 @@ public class IniFileEx: IniFile
...
+        if (File.Exists(rulesPath))
+            return new(rulesPath, ccFileManager);
+
+        var rulesBytes = ccFileManager.LoadFile(filePath);
```
Rampastring: See above.

src/TSMapEditor/Extensions/IniFileEx.cs
```diff
@@ -15,34 +16,59 @@ public class IniFileEx: IniFile
...
+        Inherit(ccFileManager);
+    }
+
+    /// <summary>
+    /// Creates an INI file from a directory + path or from MIX file system.
+    /// </summary>
+    /// <param name="filePath">Path to file or file name inside MIX file system.</param>
+    /// <param name="gameDirectory">The path to the game directory.</param>
+    /// <param name="ccFileManager">File manager object holding MIXes.</param>
+    /// <returns>Constructed INI file or empty INI file on failure.</returns>
```
Rampastring: I'd prefer the comment as `Loaded INI file, or empty INI file object if the file was not found.`. That way the success and failure cases are cleanly separated by the comma.

src/TSMapEditor/Models/Map.cs
```diff
@@ -151,6 +151,7 @@ public Rectangle LocalSize
 
         private readonly Initializer initializer;
 
+        public CCFileManager ccFileManager;
```
Rampastring: Is there any reason for this to be public? If not, make it private. If yes, then start it with an uppercase letter (and make it a property), as public field and property names in C# are started with an uppercase letter.

If it's only assigned in the constructor and never re-assigned afterwards, also make it `readonly`.

Rampastring (changes requested):

src/TSMapEditor/UI/Windows/MainMenuWindows/MapSetup.cs
```diff
@@ -31,17 +32,20 @@ public static class MapSetup
         /// <returns>Null of loading the map was successful, otherwise an error message.</returns>
         public static string InitializeMap(string gameDirectory, bool createNew, string existingMapPath, string newMapTheater, Point2D newMapSize, WindowManager windowManager)
         {
-            IniFileEx rulesIni = new(Path.Combine(gameDirectory, Constants.RulesIniPath));
-            IniFileEx firestormIni = new(Path.Combine(gameDirectory, Constants.FirestormIniPath));
-            IniFileEx artIni = new(Path.Combine(gameDirectory, Constants.ArtIniPath));
-            IniFileEx artFSIni = new(Path.Combine(gameDirectory, Constants.FirestormArtIniPath));
+            ccFileManager.GameDirectory = gameDirectory;
```
Rampastring: You need to reinitialize the `CCFileManager` instance here, otherwise you will cause potential problems if you open multiple maps in one session.

An alternative solution is adding a `Clear` method to the `CCFileManager` class that clears the search directories, mix files, and file location information, and then calling that method here.

Rampastring (changes requested):

Rampastring (approved):

Rampastring (approved):

PR #23, opened 2023-07-03
Only warn when building Palette is missing

MortonPL:
This PR adds a fallback to unit or theater palette when a BuildingType has `Palette` set but no file is found. This replicates RA2/YR and Final Alert behaviour.

Additionally, a warning will be logged.


Rampastring (changes requested):
First off, I'm not sure if I like the idea of this. I'm kind of against automatically fixing mistakes made by modders (even if the game tolerates it), as it might eventually become quite a rabbit hole trying to account for all the mistakes that mods have in their files. In general, I'd rather have the modders fix the bugs in their mods.

Secondly, I'd implement this through introducing a `GetPaletteOrDefault(string paletteFileName, Palette defaultPalette)` function as an alternative to `GetPaletteOrFail`. Do not design code that relies on exceptions if you can avoid it, as throwing and catching exceptions is expensive and usually signals that something "exceptionally bad" has happened, while that is not really the case here if there's a default to fall back to.

Rampastring (approved):

PR #24, opened 2023-07-03
Fix "Change owner" action tile highlight position

MortonPL:
This PR corrects highlight height offset of `ChangeOwnerTechnoAction` and replaces plain rectangle shape with a tile-shaped diamond.


Rampastring (approved):
Good job!

PR #26, opened 2023-07-03
Show marble madness tiles in TileDisplay when in framework mode

MortonPL:
This PR makes TileDisplay show marble madness tile graphics when the editor is in framework mode.

### DIFF
--- a/src/TSMapEditor/UI/TileDisplay.cs
+++ b/src/TSMapEditor/UI/TileDisplay.cs
@@ -62,6 +62,8 @@ namespace TSMapEditor.UI
 
         private int viewY = 0;
 
+        private bool isMarbleMadness = false;
+
         public override void Initialize()
         {
             base.Initialize();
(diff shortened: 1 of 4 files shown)


Rampastring (changes requested):
I'd prefer making `EditorState` accessible to `TileDisplay`, and adding the `OnMarbleMadnessChanged` event there. `TileDisplay` could then subscribe to that event.

In addition to introducing an additional state variable that is unnecessary, this implementation currently has the issue of what happens if the framework mode is sometimes triggered by something else than the keyboard command.

src/TSMapEditor/UI/TileDisplay.cs
```diff
@@ -171,7 +176,11 @@ private void RefreshGraphics()
                 if (tileIndex > theaterGraphics.TileCount)
                     break;
 
-                TileImage tileImage = theaterGraphics.GetTileGraphics(tileIndex);
+                TileImage tileImage;
+                if (editorState.IsMarbleMadness)
+                    tileImage = theaterGraphics.GetMarbleMadnessTileGraphics(tileIndex);
+                else
+                    tileImage = theaterGraphics.GetTileGraphics(tileIndex);
                 if (tileImage == null)
```
Rampastring, line 181: Leave an empty space between this condition and the one above.

Also, the above one could be a good place for using a ternary operator to shorten the code.

Rampastring (changes requested):
Looks better now, just stylistical issues.

Rampastring (approved):

Issue #31, opened 2023-07-08
Support light emitting BuildingTypes


MortonPL:
The map editor should display tiled light emitted from BuildingTypes.

Example configuration from `rulesmd.ini`:
```ini
[GALITE]
LightVisibility=5000
LightIntensity=0.2
LightRedTint=0.05
LightGreenTint=0.05
LightBlueTint=0.01
; rest omitted for brevity
```

ATHSE:
This should be implemented with a toggle.

Issue #30, opened 2023-07-08
Display AlphaImages attached to objects


MortonPL:
The editor should display AlphaImage graphics of objects placed on the map. Details on AlphaImage quirks: https://modenc.renegadeprojects.com/AlphaImage

Example configuration in `rulesmd.ini`:
```ini
[SOMEBUILDING]
AlphaImage=          ; file name without .shp extension
```

SCIPCION:
and TerrainTypes.

PR #35, opened 2023-07-08
Add map lighting

### LINKED ISSUE #29: Add map lighting and ion storm lighting
Currently maps are not affected by map lighting. The editor should be able to display map or ion storm lighting by tinting and darkening all graphics. The editor could also take `Ground` value into account, changing brightness of cells at different heights.

MortonPL:
Closes #29. This PR introduces an option to simulate in-game map lighting set up in the `[Lighting]` section.

### New `Constants.ini` options
```ini
; Should Tiberium be affected by map lighting?
TiberiumAffectedByLighting=false

; Should Tiberium spawning terrain objects be affected by map lighting?
TerrainObjectsAffectedByLighting=false

; Should buildings with TerrainPalette=yes be affected by map lighting?
TerrainPaletteBuildingsAffectedByLighting=false

; Should voxel models be affected by map lighting?
VoxelsAffectedByLighting=false
```

### Other config changes
* introduced a `ddLightingPreview:XNADropDown` in `Config/UI/Windows/LightingSettingsWindow.ini`, which allows the user to select a lighting preview mode:
  * Option 0 is "lighting disabled" (default),
  * Option 1 is normal map lighting
  * Option 2 is Ion Storm / Lighting Storm map lighting,
  * Option 3 is Psychic Dominator map lighting (effective in YR only),
  * Options above that have no effect and act like option 0.

### Related features and issues not addressed by this PR
* #31 Light emitting buildings, aka lamp posts,
* Support for `Ground` and `Level` value, aka tile brightness affected by cell height,
* #30 Alpha images.

### Implementation notes
* `XNAPalette` now holds two textures and two palettes, an original one and a lighting-adjusted one.
* Visibility of several palette-related properties across the project have been restricted to private to enforce respecting lighing.
* It has been assumed that every palette is affected by lighting.
* `ShapeImage`s and `VoxelModel`s store information whether they are affected by lighting,
  * All game limitations are respected by default, but can be disabled by the config maker,
    * The only inaccuracy being voxels, which in-game are not affected by tint, but are by ambient value, while here they are affected either by neither tint nor ambient (default behavior) or both.
* Remap colors are *not* tinted, but their brightness is affected by the palette, due to the current shader implementation.
* Current approach to voxel lighting requires rendered voxel frame cache invalidation on each lighting change.

### DIFF
--- a/src/TSMapEditor/Config/Constants.ini
+++ b/src/TSMapEditor/Config/Constants.ini
@@ -39,7 +39,7 @@ ReverseFacing=false
 TiberiumAffectedByLighting=false
 
 ; Should Tiberium spawning terrain objects be affected by map lighting?
-TerrainObjectsAffectedByLighting=false
+TiberiumTreesAffectedByLighting=false
 
 ; Should buildings with TerrainPalette=yes be affected by map lighting?
 TerrainPaletteBuildingsAffectedByLighting=false
--- a/src/TSMapEditor/Config/Constants.ini
+++ b/src/TSMapEditor/Config/Constants.ini
@@ -35,6 +35,15 @@ TheaterPaletteForTiberium=true
 ; Should unit/turret facings be rotating counter-clockwise?
 ReverseFacing=false
 
+; Should Tiberium be affected by map lighting?
+TiberiumAffectedByLighting=false
+
+; Should Tiberium spawning terrain objects be affected by map lighting?
+TerrainObjectsAffectedByLighting=false
+
+; Should buildings with TerrainPalette=yes be affected by map lighting?
+TerrainPaletteBuildingsAffectedByLighting=false
+
 ; The file name of the executable that the map editor expects to find from the game directory.
 ; Used for the verification that the user has given us the correct game directory.
 ExpectedClientExecutableName=DTA.exe
--- a/src/TSMapEditor/Config/Constants.ini
+++ b/src/TSMapEditor/Config/Constants.ini
@@ -44,6 +44,9 @@ TerrainObjectsAffectedByLighting=false
 ; Should buildings with TerrainPalette=yes be affected by map lighting?
 TerrainPaletteBuildingsAffectedByLighting=false
 
+; Should voxel models be affected by map lighting?
+VoxelsAffectedByLighting=false
+
 ; The file name of the executable that the map editor expects to find from the game directory.
 ; Used for the verification that the user has given us the correct game directory.
 ExpectedClientExecutableName=DTA.exe
--- a/src/TSMapEditor/Models/Lighting.cs
+++ b/src/TSMapEditor/Models/Lighting.cs
@@ -8,7 +8,7 @@ namespace TSMapEditor.Models
     {
         private const string LightingIniSectionName = "Lighting";
 
-        public event EventHandler Changed;
+        public event EventHandler ColorsRefreshed;
 
         public double Red { get; set; }
         public double Green { get; set; }
@@ -68,7 +68,7 @@ namespace TSMapEditor.Models
             IonColor = RefreshLightingColor(IonRed, IonGreen, IonBlue, IonAmbient);
             DominatorColor = RefreshLightingColor(DominatorRed, DominatorGreen, DominatorBlue, DominatorAmbient);
 
-            Changed?.Invoke(this, EventArgs.Empty);
+            ColorsRefreshed?.Invoke(this, EventArgs.Empty);
         }
 
         private static Color RefreshLightingColor(double? red, double? green, double? blue, double? ambient)
--- a/src/TSMapEditor/Models/Lighting.cs
+++ b/src/TSMapEditor/Models/Lighting.cs
@@ -1,10 +1,14 @@
-﻿using Rampastring.Tools;
+﻿using Microsoft.Xna.Framework;
+using Rampastring.Tools;
+using System;
 
 namespace TSMapEditor.Models
 {
     public class Lighting : INIDefineable
     {
         private const string LightingIniSectionName = "Lighting";
+
+        public event EventHandler Changed;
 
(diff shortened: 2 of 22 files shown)


src/TSMapEditor/Models/Lighting.cs
```diff
@@ -49,5 +58,22 @@ public void WriteToIniFile(IniFile iniFile)
...
+            IonColor = RefreshLightingColor(IonRed, IonGreen, IonBlue, IonAmbient);
+            DominatorColor = RefreshLightingColor(DominatorRed, DominatorGreen, DominatorBlue, DominatorAmbient);
+
+            Changed?.Invoke(this, EventArgs.Empty);
+        }
+
+        private static Color RefreshLightingColor(double? red, double? green, double? blue, double? ambient)
+        {
+            if (red == null || green == null || blue == null || ambient == null)
+                return Color.White;
```
ZivDero, line 77: Could this result in Dominator lighting being set to white in a TS map and then written to the map pointlessly?

MortonPL: `NormalColor`, `IonColor` and `DominatorColor` should not be serialized now.

src/TSMapEditor/Config/Constants.ini
```diff
@@ -35,6 +35,18 @@ TheaterPaletteForTiberium=true
 ; Should unit/turret facings be rotating counter-clockwise?
 ReverseFacing=false
 
+; Should Tiberium be affected by map lighting?
+TiberiumAffectedByLighting=false
+
+; Should Tiberium spawning terrain objects be affected by map lighting?
+TerrainObjectsAffectedByLighting=false
```
Rampastring: `TerrainObjectsAffectedByLighting` sounds like it'd affect all terrain objects, while the comment says that this key would only affect Tiberium-spawning terrain objects. Give the INI key a name that matches its purpose better.

src/TSMapEditor/Models/Lighting.cs
```diff
@@ -1,11 +1,15 @@
...
+using Rampastring.Tools;
+using System;
 
 namespace TSMapEditor.Models
 {
     public class Lighting : INIDefineable
     {
         private const string LightingIniSectionName = "Lighting";
 
+        public event EventHandler Changed;
```
Rampastring: Reading the code below, a more appropriate name would be `ColorsRefreshed` than `Changed`. I'd expect a `Changed` event to trigger when the individual color components are changed.

src/TSMapEditor/Rendering/TheaterGraphics.cs
```diff
@@ -1270,12 +1300,31 @@ public void ReadSmudgeTextures(List<SmudgeType> smudgeTypes)
...
 
             Logger.Log("Finished loading smudge textures.");
         }
 
+        public void ApplyLightingToPalettes(Color lighting)
+        {
+            theaterPalette.ApplyLighting(lighting);
+            unitPalette.ApplyLighting(lighting);
+            tiberiumPalette.ApplyLighting(lighting);
+            animPalette.ApplyLighting(lighting);
```
Rampastring: The `palettes` list also contains these standard palettes, so these lines are unnecessary.

src/TSMapEditor/Rendering/TheaterGraphics.cs
```diff
@@ -155,19 +159,26 @@ public PositionedTexture GetRemapFrame(byte facing, RampType ramp)
             }
         }
 
+        public void ClearFrames()
+        {
+            Frames.Clear();
+            RemapFrames.Clear();
```
Rampastring, line 171: You need to dispose all `Texture2D` instances within these or you will be leaking GPU memory.

src/TSMapEditor/UI/Windows/LightingSettingsWindow.cs
```diff
@@ -125,6 +142,17 @@ private void BtnApply_LeftClick(object sender, System.EventArgs e)
...
             }
 
+            state.LightingPreviewState = ddLightingPreview.SelectedIndex switch
+            {
+                0 => LightingPreviewMode.NoLighting,
+                1 => LightingPreviewMode.Normal,
+                2 => LightingPreviewMode.IonStorm,
+                3 => LightingPreviewMode.Dominator,
+                _ => LightingPreviewMode.NoLighting,
+            };
```
Rampastring: You could add numbers to the values of the `LightPreviewMode` enum so this could be simplified to `state.LightingPreviewState = (LightingPreviewMode)ddLightingPreview.SelectedIndex.

The lines beween 108 and 112 could then also be similarly replaced.

Rampastring (changes requested):
Looks mostly good, there's a few things to fix or improve.

src/TSMapEditor/Config/Constants.ini
ZivDero: TerrainObjects -> Tibtrees?

src/TSMapEditor/UI/Windows/LightingSettingsWindow.cs
MortonPL: I disagree, these enum variants shouldn't have assigned integer values.
Things like `state.LightingPreviewState = (LightingPreviewMode)99` should not be allowed.

Rampastring: I'd agree if this was somehow decoupled from indexes in general, but it's already tied to the selected index of `ddLightingPreview`. If you want to avoid stuff like `(LightingPreviewMode)99`, you could check the value with `Enum.IsDefined` and fall back to `LightingPreviewMode.NoLighting` if the value is nonsensical. You could do it pretty cleanly in the setter in `EditorState.LightingPreviewState`.

Rampastring (approved):

PR #36, opened 2023-07-08
Example configuration for YR

MortonPL:
This PR provides an example configuration for Yuri's Revenge and YR-based modifications. It's targeting branch `tsclient` for now to minimize diff, but, once complete, it should ideally become an upstream branch.

Current configuration contains most of the necessary map making things, but it is currently missing mission scripting data like new trigger events, actions and script actions.


ZivDero:
Removed all the code changes from this branch, now it only consists of config changes. (Except for the last few commits, consider splitting them).
Rebased onto the latest TS config.

src/TSMapEditor/Config/Actions.ini
```diff
@@ -134,7 +134,7 @@ Description=Reveals all cells that share the same zone as the waypoint specified
 P2Type=Waypoint
 
 [PlaySound]
-Name=Play sound
+Name=Play Sound
```
Rampastring: I'd like both trigger/script action names and descriptions to be uniform across editor configurations, for the actions that are shared between TS and YR. If you feel they can be improved, make a PR for it targeting `master` rather than editing them for YR alone.

Rampastring (changes requested):

src/TSMapEditor/Config/Actions.ini
ZivDero: > I'd like both trigger/script action names and descriptions to be uniform across editor configurations, for the actions that are shared between TS and YR. If you feel they can be improved, make a PR for it targeting `master` rather than editing them for YR alone.

I totally agree. Will need to put some thought into stadardizing it, though, as rn it's a mish-mash of FA2/FS, ModEnc and "freestyle".

Rampastring (approved):

PR #42, opened 2023-08-04
Update to GPL version 3.0

Mailaender:
@OpenRA is GPLv3 that makes your project GPLv3 as well. https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility You also need to include a full copy by the way.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -6,7 +6,7 @@ https://www.moddb.com/mods/the-dawn-of-the-tiberium-age
 
 This program is free software; you can redistribute it and/or
 modify it under the terms of the GNU General Public License
-as published by the Free Software Foundation; either version 2
+as published by the Free Software Foundation; either version 3
 of the License, or (at your option) any later version.
 
 This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


LICENSE.txt
```diff
@@ -22,8 +22,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 This program uses code from the OpenRA project. It is made
 available to you under the terms of the GNU General Public License
-as published by the Free Software Foundation.
-Copyright 2007-2011 The OpenRA Developers
+version 3 as published by the Free Software Foundation.
+Copyright 2007-2023 The OpenRA Developers
```
Rampastring: The only file from OpenRA used is Format80, from 2011. Thus I'm not sure if it's necessary or appropriate to update the OpenRA license year in the license file to 2023.

If you think it's still better as 2023, I can accept the change though.

Mailaender: [converted to GPLv3](https://github.com/OpenRA/OpenRA/commit/a72d19e754e049c75d8e73d7a9b502b09303d308) was 2010

Mailaender: I thought more files were used.

Rampastring (approved):

PR #46, opened 2023-08-09
Fix NewTheater=no buildings not loading when generic version is missing

ZivDero:
Buildings that use NewTheater in RA2/YR but for some reason don't have generic art now fall back to the theater-correct version, and not the arctic version, which led to them appearing without art.

### DIFF
--- a/src/TSMapEditor/Rendering/TheaterGraphics.cs
+++ b/src/TSMapEditor/Rendering/TheaterGraphics.cs
@@ -637,7 +637,9 @@ namespace TSMapEditor.Rendering
                 // if the theater-specific SHP is not found
                 if (shpData == null)
                 {
-                    shpData = fileManager.LoadFile(shpFileName);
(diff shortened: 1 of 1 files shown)


Rampastring:
Does this happen for any vanilla RA2/YR buildings? If it's only an issue in mods, I'd rather not fix this, for the same reasons as described here: https://github.com/Rampastring/TSMapEditor/pull/23

ZivDero:
It does happen in RA2/YR. I noticed it with flag buildings, for example CARUFGL. CGRUFGL doesn't exist even though it's NewTheater=no. YR displays it properly showing CTRUFGL, while WAE tried to load CARUFGL and didn't find it, of course.

Rampastring:
This should be made obsolete by https://github.com/Rampastring/TSMapEditor/commit/3a846a61764f7f40d39970102e1fcb6fb1630c66

PR #47, opened 2023-08-09
Add EditorRules.ini sections for Category overrides and Ignores

ZivDero:
Previously, it was only possible to set the category for a specific Techno, or hide it from the editor only in rules.ini. This PR adds sections in EditorRules.ini for those who'd prefer doing it in the editor's settings.

### DIFF
--- a/src/TSMapEditor/Config/EditorRules.ini
+++ b/src/TSMapEditor/Config/EditorRules.ini
@@ -10,7 +10,7 @@ RenderInEditor=no
 [RAAPWR_S]
 RenderInEditor=no
 
-[ObjectCategoryOverrides]
+[ObjectOwnerOverrides]
 GDI1=GDI
 GDI2=GDI
 Nod1=Nod
@@ -19,6 +19,10 @@ Allies1=Allies
 Allies2=Allies
 Soviet1=Soviet
 Soviet2=Soviet
+
+[ObjectCategoryOverrides]
+
+[IgnoreTypes]
 
 [ObjectCategoryPriorities]
 GDI=0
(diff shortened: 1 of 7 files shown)


Rampastring (changes requested):
I'm not a fan of accessing the INI data when initializing objects for the sidebar. In such a case, an object might have `EditorVisible=true` but still not get displayed in the sidebar, which'd be confusing.

A better way would be adding more editor rules init logic that, after initializing rules for the map, would go through the ignored types and give them `EditorVisible = false`.

Further, instead of just checking whether the key exists, I'd use `GetBooleanValue`, so you'd for example specify it as

```ini
[IgnoreTypes]
CABHUT=yes
```

and could then disable the ignore by setting CABHUT to false. As it is, the value of the key would not matter at all, which would be unusual for INI data.

ZivDero:
I've updated it to use GetBooleanValue.

I'm not qute sure how you meant to have EditorVisible=false set, since when the types are loaded EditorVisible gets defaulted to true if it doesn't exist, so once they are loaded we can't check which ones had it set in the ini and which ones didn't.

What I suppose could be done is EditorVisible could be turned into bool?, and then we can indeed do that.

Rampastring:
This is probably made obsolete by https://github.com/Rampastring/TSMapEditor/commit/b1388f5a7e35703fce51da84efcaf4f8de374028 ?

ZivDero:
> This is probably made obsolete by [b1388f5](https://github.com/Rampastring/TSMapEditor/commit/b1388f5a7e35703fce51da84efcaf4f8de374028) ?

Not really. For YR I currently have 148 ignores and 133 category overrides. Putting them all in as sections would effectively triple the space it takes (since instead of a single Key=Value, you need [Section], then Key=Value, and also an empty line). So this solution is extremely inefficient.

Mind that of all the category changes, only 8 are a "custom" category of Tech buildings, the rest are just fixing Allies, Soviets and Yuri as Owner= is not very representative in YR and cannot be relied upon.

ZivDero:
I have updated the PR to have all the overrides loaded in the beginning.

Rampastring (approved):

PR #48, opened 2023-08-09
Allow multiple objects in cell

ZivDero:
This PR converts Waypoint, Aircraft, Vehicle and Structure into Lists, allowing to place multiple of those entities in a single cell, which is supported by the game engine.

### DIFF
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -352,8 +352,6 @@ namespace TSMapEditor.Initialization
 
                     if (tile.Structures.Count > 0)
                     {
-                        //isClear = false;
-                        //AddMapLoadError($"Building {buildingType.ININame} exists on a cell ({cellCoords}) that already has another building ({tile.Structure.ObjectType.ININame}). Skipping adding it to map.");
                         Logger.Log($"NOTE: Building {buildingType.ININame} exists in the cell {cellCoords} that already has other buildings.");
                     }
                 }
@@ -677,7 +675,6 @@ namespace TSMapEditor.Initialization
 
                 if (mapCell.Waypoints.Count > 0)
                 {
-                    //AddMapLoadError($"Cell at {waypoint.Position} has multiple waypoints placed on it. Skipping adding waypoint #{waypoint.Identifier} there.");
                     Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists in the cell {waypoint.Position} that already has other waypoints.");
                     continue;
                 }
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -4,6 +4,7 @@ using Rampastring.Tools;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Text;
 using TSMapEditor.GameMath;
 using TSMapEditor.Models;
@@ -352,7 +353,7 @@ namespace TSMapEditor.Initialization
 
                     if (tile.Structures.Count > 0)
                     {
-                        Logger.Log($"NOTE: Building {buildingType.ININame} exists in the cell {cellCoords} that already has other buildings.");
+                        Logger.Log($"NOTE: Building {buildingType.ININame} exists in the cell at {cellCoords} that already has other buildings: {string.Join(", ", tile.Structures.Select(s => s.ObjectType.ININame))}");
                     }
                 }
 
@@ -666,16 +667,16 @@ namespace TSMapEditor.Initialization
                     continue;
                 }
 
-                var mapCell = map.GetTile(waypoint.Position.X, waypoint.Position.Y);
-                if (mapCell == null)
+                var tile = map.GetTile(waypoint.Position.X, waypoint.Position.Y);
+                if (tile == null)
                 {
                     AddMapLoadError($"Waypoint {waypoint.Identifier} at {waypoint.Position} is not within the valid map area.");
                     continue;
                 }
 
-                if (mapCell.Waypoints.Count > 0)
+                if (tile.Waypoints.Count > 0)
                 {
-                    Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists in the cell {waypoint.Position} that already has other waypoints.");
+                    Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists in the cell at {waypoint.Position} that already contains other waypoints: {string.Join(", ", tile.Waypoints.Select(s => s.Identifier))}");
                     continue;
                 }
 
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -350,10 +350,11 @@ namespace TSMapEditor.Initialization
                         return;
                     }
 
-                    if (tile.Structure != null)
+                    if (tile.Structures.Count > 0)
                     {
-                        isClear = false;
-                        AddMapLoadError($"Building {buildingType.ININame} exists on a cell ({cellCoords}) that already has another building ({tile.Structure.ObjectType.ININame}). Skipping adding it to map.");
+                        //isClear = false;
+                        //AddMapLoadError($"Building {buildingType.ININame} exists on a cell ({cellCoords}) that already has another building ({tile.Structure.ObjectType.ININame}). Skipping adding it to map.");
+                        Logger.Log($"NOTE: Building {buildingType.ININame} exists on the cell {cellCoords} that already has other buildings.");
                     }
                 }
 
@@ -366,7 +367,7 @@ namespace TSMapEditor.Initialization
                 buildingType.ArtConfig.DoForFoundationCoordsOrOrigin(offset =>
                 {
                     var tile = map.GetTile(building.Position + offset);
-                    tile.Structure = building;
+                    tile.Structures.Add(building);
                 });
             }
         }
@@ -673,12 +674,13 @@ namespace TSMapEditor.Initialization
                     AddMapLoadError($"Waypoint {waypoint.Identifier} at {waypoint.Position} is not within the valid map area.");
                     continue;
                 }
-                
-                //if (mapCell.Waypoint != null)
-                //{
-                //    AddMapLoadError($"Cell at {waypoint.Position} has multiple waypoints placed on it. Skipping adding waypoint #{waypoint.Identifier} there.");
-                //    continue;
-                //}
+
+                if (mapCell.Waypoints.Count > 0)
+                {
+                    //AddMapLoadError($"Cell at {waypoint.Position} has multiple waypoints placed on it. Skipping adding waypoint #{waypoint.Identifier} there.");
+                    Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists on the cell {waypoint.Position} that already has other waypoints.");
+                    continue;
+                }
 
                 map.AddWaypoint(waypoint);
             }
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -354,7 +354,7 @@ namespace TSMapEditor.Initialization
                     {
                         //isClear = false;
                         //AddMapLoadError($"Building {buildingType.ININame} exists on a cell ({cellCoords}) that already has another building ({tile.Structure.ObjectType.ININame}). Skipping adding it to map.");
-                        Logger.Log($"NOTE: Building {buildingType.ININame} exists on the cell {cellCoords} that already has other buildings.");
+                        Logger.Log($"NOTE: Building {buildingType.ININame} exists in the cell {cellCoords} that already has other buildings.");
                     }
                 }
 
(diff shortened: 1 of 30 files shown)


Rampastring (changes requested):
The game does not fully support this functionality. In particular, if buildings placed on top of each other, the game crashes if one of them is destroyed. This was a common mapping error with FinalSun/FinalAlert, which WAE was designed to prevent.

There are still a couple of good reasons for in turning a cell's buildings into a list though:

1) it has use for buildings that are converted to overlay. For example, placing a crate building with `ToOverlay=CRATE` under a real destroyable building, to add a collectible crate that the player can pick up after destroying the building
2) supporting maps that already have overlapping buildings

However I don't think WAE should allow placing overlapping buildings by default. Instead this should be behind some modifier key, so the mapper can't accidentally make buildings overlap, but instead it's always a deliberate action.

For units, aircraft and waypoints there is no such issue AFAIK, but is it _beneficial_ to allow them to overlap? I have a hard time imagining a case where making units overlap would be very useful. It could still be allowed with the modifier key, but I'd also prevent it by default.

For the implementation, I think the current implementation might crash when deleting objects as you're modifying a collection that is being foreach'd. Deletion should instead act on a copy of the list.

ZivDero:
YR does in fact fully support this and does not crash (whether the buildings are invincible or not). Although it seems that invincile buildings also get damaged when the vulnerable ones receive damage. I suppose adding a modifyer key wouldn't hurt, though.

As for deletion, in my tests it did not crash but I agree that this is not a safe way to do it indeed so will redo.
EDIT: In fact, it should be safe as I'm not performing actions on the collecttion that's being foreached, rather on a different collection.

ZivDero:
Okay, all done. I've added Alt as the default hotkey that allows to place overlapping objects. Works for placing new buildings, moving/cloning buildings, as well as the Terrain copy tool. Don't think I've missed anything.

src/TSMapEditor/Models/MapTile.cs
```diff
@@ -68,11 +67,11 @@ public void ShiftPosition(int x, int y)
...
         public void AddObjectsToList(List<AbstractObject> objects)
         {
-            if (Structure != null)
-                objects.Add(Structure);
+            if (Structures.Count > 0)
+                objects.AddRange(Structures);
 
-            if (Vehicle != null)
-                objects.Add(Vehicle);
+            if (Structures.Count > 0)
```
Rampastring: Looks like a copy/paste error.

src/TSMapEditor/Mutations/Classes/PasteTerrainMutation.cs
```diff
@@ -616,8 +620,19 @@ public override void Perform()
                 if (cell == null)
                     continue;
 
-                if (cell.Vehicle != null)
-                    continue;
+                if (cell.Vehicles.Count > 0 && !allowOverlap)
+                {
+                    bool foreignUnitExists = false;
+                    foreach (var vehicle in cell.Vehicles)
```
Rampastring, line 626: Only statements at the most indented level can be braceless. Iow. add braces to this `foreach`.

src/TSMapEditor/Mutations/Classes/PasteTerrainMutation.cs
```diff
@@ -657,31 +675,43 @@ public override void Perform()
...
-
-                if (!isFoundationClear)
-                    continue;
+                    bool isFoundationClear = true;
+                    buildingType.ArtConfig.DoForFoundationCoords(offset =>
+                    {
+                        Point2D foundationCellCoords = offset + cellCoords;
+                        MapTile foundationCell = MutationTarget.Map.GetTile(foundationCellCoords);
+
+                        foreach (var building in foundationCell.Structures)
```
Rampastring, line 692: Only statements at the most indented level can be braceless. Iow. add braces to this `foreach`.

src/TSMapEditor/UI/CursorActions/AircraftPlacementAction.cs
```diff
@@ -8,32 +9,35 @@ namespace TSMapEditor.UI.CursorActions
...
+        public AircraftPlacementAction(ICursorActionTarget cursorActionTarget, RKeyboard keyboard) : base(cursorActionTarget)
         {
+            this.keyboard = keyboard;
         }
 
         public override string GetName() => "Place Aircraft";
 
         private Aircraft aircraft;
 
-        private AircraftType _aircraftType;
```
Rampastring, line 19: Keep the underscore in names of fields that only exist to back a property.

src/TSMapEditor/UI/CursorActions/AircraftPlacementAction.cs
```diff
@@ -44,43 +48,53 @@ public override void PreMapDraw(Point2D cellCoords)
...
-            if (tile.Aircraft == null)
-            {
-                tile.Aircraft = aircraft;
-                CursorActionTarget.TechnoUnderCursor = aircraft;
-                CursorActionTarget.AddRefreshPoint(cellCoords);
-            }
+            tile.Aircraft.Add(aircraft);
+            CursorActionTarget.TechnoUnderCursor = aircraft;
+            CursorActionTarget.AddRefreshPoint(cellCoords);
+
```
Rampastring: Pointless blank line.

src/TSMapEditor/UI/CursorActions/AircraftPlacementAction.cs
```diff
@@ -44,43 +48,53 @@ public override void PreMapDraw(Point2D cellCoords)
             // Assign preview data
             aircraft.Position = cellCoords;
 
+            bool overlapObjects = KeyboardCommands.Instance.OverlapObjects.AreKeysOrModifiersDown(keyboard);
+
+            bool canPlace = Map.CanPlaceObjectAt(aircraft, cellCoords, false,
+                overlapObjects);
```
Rampastring: This line isn't long enough to warrant cutting it.

src/TSMapEditor/UI/CursorActions/BuildingPlacementAction.cs
```diff
@@ -11,32 +13,35 @@ namespace TSMapEditor.UI.CursorActions
...
+        public BuildingPlacementAction(ICursorActionTarget cursorActionTarget, RKeyboard keyboard) : base(cursorActionTarget)
         {
+            this.keyboard = keyboard;
         }
 
         public override string GetName() => "Place Building";
 
         private Structure structure;
 
-        private BuildingType _buildingType;
```
Rampastring, line 22: Keep the underscore in names of fields that only exist to back a property.

src/TSMapEditor/UI/CursorActions/InfantryPlacementAction.cs
```diff
@@ -19,25 +19,25 @@ public InfantryPlacementAction(ICursorActionTarget cursorActionTarget) : base(cu
 
         private Infantry infantry;
 
-        private InfantryType _infantryType;
```
Rampastring: Keep the underscore in names of fields that only exist to back a property.

src/TSMapEditor/UI/CursorActions/OverlayCollectionPlacementAction.cs
```diff
@@ -15,18 +15,18 @@ public OverlayCollectionPlacementAction(ICursorActionTarget cursorActionTarget)
 
         public override string GetName() => "Place Overlay Collection";
 
-        private OverlayCollection _overlayCollection;
```
Rampastring: Keep the underscore in names of fields that only exist to back a property.

src/TSMapEditor/UI/CursorActions/OverlayPlacementAction.cs
```diff
@@ -20,15 +20,15 @@ public OverlayPlacementAction(ICursorActionTarget cursorActionTarget) : base(cur
 
         public event EventHandler OverlayTypeChanged;
 
-        private OverlayType _overlayType;
```
Rampastring: Keep the underscore in names of fields that only exist to back a property.

src/TSMapEditor/UI/CursorActions/TerrainObjectPlacementAction.cs
```diff
@@ -19,24 +19,24 @@ public TerrainObjectPlacementAction(ICursorActionTarget cursorActionTarget) : ba
 
         private TerrainObject terrainObject;
 
-        private TerrainType _terrainType;
```
Rampastring: Keep the underscore in names of fields that only exist to back a property.

src/TSMapEditor/UI/CursorActions/UnitPlacementAction.cs
```diff
@@ -11,32 +12,35 @@ namespace TSMapEditor.UI.CursorActions
...
+        public UnitPlacementAction(ICursorActionTarget cursorActionTarget, RKeyboard keyboard) : base(cursorActionTarget)
         {
+            this.keyboard = keyboard;
         }
 
         public override string GetName() => "Place Vehicle";
 
         private Unit unit;
 
-        private UnitType _unitType;
```
Rampastring, line 22: Keep the underscore in names of fields that only exist to back a property.

src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -350,10 +350,9 @@ void CheckFoundationCell(Point2D cellCoords)
                         return;
                     }
 
-                    if (tile.Structure != null)
+                    if (tile.Structures.Count > 0)
                     {
-                        isClear = false;
-                        AddMapLoadError($"Building {buildingType.ININame} exists on a cell ({cellCoords}) that already has another building ({tile.Structure.ObjectType.ININame}). Skipping adding it to map.");
+                        Logger.Log($"NOTE: Building {buildingType.ININame} exists in the cell {cellCoords} that already has other buildings.");
```
Rampastring: Can improve this a bit (both grammar and accuracy wise).

```suggestion
                        Logger.Log($"NOTE: Building {buildingType.ININame} exists in a cell {cellCoords} that already has one or more buildings.");
```

src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -673,10 +672,10 @@ public static void ReadWaypoints(IMap map, IniFile mapIni)
                     AddMapLoadError($"Waypoint {waypoint.Identifier} at {waypoint.Position} is not within the valid map area.");
                     continue;
                 }
-                
-                if (mapCell.Waypoint != null)
+
+                if (mapCell.Waypoints.Count > 0)
                 {
-                    AddMapLoadError($"Cell at {waypoint.Position} has multiple waypoints placed on it. Skipping adding waypoint #{waypoint.Identifier} there.");
+                    Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists in the cell {waypoint.Position} that already has other waypoints.");
```
Rampastring: Can improve this a bit (both grammar and accuracy wise).
```suggestion
                    Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists in a cell {waypoint.Position} that already has one or more waypoints.");
```

src/TSMapEditor/Models/Map.cs
```diff
@@ -597,31 +597,34 @@ public void AddWaypoint(Waypoint waypoint)
         {
             Waypoints.Add(waypoint);
             var cell = GetTile(waypoint.Position.X, waypoint.Position.Y);
-            if (cell.Waypoint != null)
+            if (cell.Waypoints.Count > 0)
             {
-                throw new InvalidOperationException($"Cell at {cell.CoordsToPoint()} already has a waypoint, skipping adding waypoint {waypoint.Identifier}");
+                Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists in the cell {cell.CoordsToPoint()} that already has other waypoints.");
```
Rampastring: Can improve this a bit (both grammar and accuracy wise).
```suggestion
                Logger.Log($"NOTE: Waypoint {waypoint.Identifier} exists in a cell {waypoint.Position} that already has one or more waypoints.");
```

Rampastring (changes requested):
Functionality looks good now.

There are some stylistical things to fix, including some harmful renames.

src/TSMapEditor/UI/CursorActions/UnitPlacementAction.cs
ZivDero: The underscores actually confused me since from what I could tell, the C# convention was to use _camelCase for all private fields, not just ones that back properties. So it seemed reasonable to either use it everywhere (which I didn't see happen much), or not use it at all. I guess what you said also can make sense?

ZivDero:
Implemented the requested changes.

src/TSMapEditor/UI/CursorActions/TerrainObjectCollectionPlacementAction.cs
```diff
@@ -15,19 +15,19 @@ public TerrainObjectCollectionPlacementAction(ICursorActionTarget cursorActionTa
         public override string GetName() => "Place TerrainObject Collection";
 
         private TerrainObject terrainObject;
-        private TerrainObjectCollection _terrainObjectCollection;
+        private TerrainObjectCollection terrainObjectCollection;
```
Rampastring: Here is one more removed underscore.

Rampastring (approved):

PR #49, opened 2023-08-09
Change waypoints outside of A-ZZ crashing the editor

ZivDero:
This PR makes it so that when the editor enounters a waypoint outside of the A-ZZ range, it doesn't throw an exception, but logs it and replaces the waypoint with 0.

### DIFF
--- a/src/TSMapEditor/Helpers.cs
+++ b/src/TSMapEditor/Helpers.cs
@@ -123,10 +123,18 @@ namespace TSMapEditor
             return landType == 0x9;
         }
 
-        public static int GetWaypointNumberFromAlphabeticalString(string str)
+        public static bool IsValidAlphabeticalWaypoint(string str)
         {
             if (str.Length < 1 || str.Length > 2 ||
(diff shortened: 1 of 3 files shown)


Rampastring (changes requested):
A better approach is to fix any cases where the current implementation might cause an issue, and never call `GetWaypointNumberFromAlphabeticalString` in the first place. I've only heard of it being an issue for TeamTypes that have no waypoint specified at all, which should be fixed in the waypoint parsing code.

In general, WAE's codebase uses a "fail fast" principle, where if something unexpected happens, instead of trying to continue operation and possibly cause weird behaviour in the process (in this case, replacing waypoint values as a surprise for the user), an exception is thrown. This makes it possible to catch mistakes and bugs early, before they cause serious issues.

ZivDero:
Made obsolete by [517b27b](https://github.com/Rampastring/TSMapEditor/commit/517b27bc30c885fdfa1418bb318e6cdd3012f15c)

PR #50, opened 2023-08-09
Add facing selection to TypeOptions

ZivDero:
This PR adds a facing dropdown to Infantry, Unit and Building Options dialogues. As a side-effect, it also fixes the bug where the buildings set to be damaged did not update their art immediately as it now updates on accepting changes.


Rampastring (commented):
Code looks OK for the most part. Could leave the size parameter out from the refreshes.

I wonder whether this is necessary though as WAE allows you to rotate units by dragging, which is more flexible and efficient than changing the facing through the properties window. Instead of only replicating FinalSun/FinalAlert functionality, WAE also aims to make workflows more efficient.

This would still be useful for a potential use case where you'd be working on a map without having or using a keyboard though, so I'll leave it for now and we can see whether mappers at large want to have this feature.

PR #54, opened 2023-08-09
Add ID & Facing info in TileInfoDisplay for other Types

ZivDero:
This PR adds ID and Facing fields to the TileInfoDisplay for the Structure, Aircraft and Infantry types.


Rampastring:
Facing information is useful. I wonder if the ID has any use though? If not, it would be better to get rid of it to keep the tile info display as minimalistic as possible.

ZivDero:
The only use I managed to come up for the ID is when you are editing the map in text and could use it, perhaps.

Issue #40, opened 2023-07-09
Dehardcode editor themes


MortonPL:
Currently there are two themes available ("Default" and "Tiberium"), neither of which is editable via INIs. It would be nice to have a system similar to that of XNA Client.

Rampastring:
Implemented in https://github.com/Rampastring/TSMapEditor/commit/be4bf98a86d99520a29a0f9378368332df959a02 and https://github.com/Rampastring/TSMapEditor/commit/8d9ff3687604e0051fcc9c50f522808d2ec5b631

PR #63, opened 2023-09-03
Dehardcoded themes

ZivDero:
This PR dehardcodes the editor themes. Themes can be customized in EditorThemes.ini.
The light theme is still the stub found in the code originally.

This PR closes issue #40.


src/TSMapEditor/UI/UIManager.cs
```diff
@@ -21,10 +24,75 @@ namespace TSMapEditor.UI
 {
     public class CustomUISettings : UISettings
```
Rampastring, line 22: I'd prefer splitting the editor themes alongside the `InitializeEditorThemes` function into a new, static `EditorThemes` class that would belong in its own file. Then there would be a clear separation between the "list of themes" and a single theme.

src/TSMapEditor/UI/UIManager.cs
```diff
@@ -21,10 +24,75 @@ namespace TSMapEditor.UI
...
+                        continue;
+
+                    if (property.PropertyType == typeof(Color))
+                    {
+                        string propertyString = themeSection.GetStringValue(property.Name, "0,0,0,255");
+                        var parts = propertyString.Split(',');
+                        if (parts.Length != 4)
+                            continue;
+
+                        property.SetValue(theme, new Color(int.Parse(parts[0]), int.Parse(parts[1]), int.Parse(parts[2]), int.Parse(parts[3])), null);
```
Rampastring: This might be used more commonly.

Add a new method to `Helpers.cs` and call it here. The function can be similar to this (taken from a private project of mine):

```cs
    public static Color ColorFromString(string str)
    {
        string[] parts = str.Split(',');

        if (parts.Length < 3 || parts.Length > 4)
            throw new ArgumentException("AColor.FromString: parameter was not in a valid format: " + str);

        int r = Conversions.IntFromString(parts[0], 0);
        int g = Conversions.IntFromString(parts[1], 0);
        int b = Conversions.IntFromString(parts[2], 0);
        int a = 255;

        if (parts.Length == 4)
            a = Conversions.IntFromString(parts[3], a);

        return new Color((byte)r, (byte)g, (byte)b, (byte)a);
    }
```

As a bonus, it also supports both 3-value and 4-value colors, making the alpha parameter optional.

Rampastring (changes requested):
Looks mostly good. There's one class design issue and one implementation that could be extracted into a helper function.

Rampastring (approved):

PR #64, opened 2023-09-03
Use WinForms for Open and Save As dialogs (in main menu and toolbar)

ZivDero:
This PR adds "Browse..." buttons to the main menu that open a WinForms dialog to select a file. It also replaces the dialogs opened by the Open and Save as buttons of the tool bar with WinForms dialogs.

### DIFF
--- a/src/TSMapEditor/Config/Constants.ini
+++ b/src/TSMapEditor/Config/Constants.ini
@@ -45,6 +45,10 @@ ExpectedClientExecutableName=DTA.exe
 ; The CnCNet Client has been programmed to write its installation path to the registry by default.
 GameRegistryInstallPath=SOFTWARE\DawnOfTheTiberiumAge
 
+; Specifies the file filter string used in OpenFileDialog (Windows)
+; For multiple file formats, list them with : (instead of ;)
+OpenFileDialogFilter=YR maps|*.map:*.yrm|All files|*.*
+
 ; Should the editor consider Ares [#include] or Phobos [$Include] section?
 EnableIniInclude=false
 ; Should the editor consider Phobos $Inherits entries?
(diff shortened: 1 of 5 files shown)


src/TSMapEditor/UI/MainMenu.cs
```diff
@@ -68,10 +85,20 @@ public override void Initialize()
...
             tbMapPath.X = Constants.UIEmptySideSpace;
             tbMapPath.Y = lblMapPath.Bottom + Constants.UIVerticalSpacing;
-            tbMapPath.Width = Width - Constants.UIEmptySideSpace * 2;
+            tbMapPath.Width = Width - Constants.UIEmptySideSpace * 3 - 70;
             tbMapPath.Text = UserSettings.Instance.LastScenarioPath;
             AddChild(tbMapPath);
 
+            btnBrowseMapPath = new EditorButton(WindowManager);
+            btnBrowseMapPath.Name = nameof(btnBrowseMapPath);
+            btnBrowseMapPath.Width = 70;
```
Rampastring: Seeing that this `70` appears 4 times here in different places, define it as a `const` at the beginning of the class. `const int BrowseButtonWidth = 70;` or something along those lines.

src/TSMapEditor/Config/Constants.ini
```diff
@@ -45,6 +45,10 @@ ExpectedClientExecutableName=DTA.exe
 ; The CnCNet Client has been programmed to write its installation path to the registry by default.
 GameRegistryInstallPath=SOFTWARE\DawnOfTheTiberiumAge
 
+; Specifies the file filter string used in OpenFileDialog (Windows)
+; For multiple file formats, list them with : (instead of ;)
```
Rampastring, line 49: Add an example of multiple file formats here in a comment so regular modders won't need to start googling for the format used in the MS API. You can for example make a listing of both *.map and *.mpr files (as the original game used both).

Rampastring (changes requested):
Looks mostly good, a couple small things to fix.

Rampastring (approved):

PR #66, opened 2023-09-18
Replaced SHP decompression with Nyerguds's algorithm

ZivDero:
The old implementation caused some compressed shapes to not be loaded correctly.


Rampastring:
Made obsolete by https://github.com/Rampastring/TSMapEditor/commit/67700a3c5bdee461f589d05f42e2bc883ebe714b

PR #70, opened 2023-09-30
Add bridge drawing tool

ZivDero:
This PR expands the place bridge functionality to a fully customizable bridge drawing tool, supporting both low and high bridges.
It also fixes the high bridges to be rendered at their correct height.


src/TSMapEditor/Config/UI/Windows/SelectBridgeWindow.ini
```diff
@@ -0,0 +1,32 @@
+[SelectBridgeWindow]
+$Width=250
+$Height=RESOLUTION_HEIGHT - 100
```
Rampastring: This is probably comically big for a list that will most of the time have only like 3 entries.

src/TSMapEditor/Models/Bridge.cs
```diff
@@ -0,0 +1,92 @@
...
+                    throw new BridgeLoadException($"Low bridge {bridge.Name} has no bridge pieces!");
+            }
+            else
+            {
+                string piece = iniSection.GetStringValue($"BridgePieces.{suffix}", null);
+                if (piece == null)
+                    throw new BridgeLoadException($"High bridge {bridge.Name} has no bridge piece!");
+                int bridgePiece = rules.FindOverlayType(piece)?.Index ??
+                      throw new BridgeLoadException($"High bridge {bridge.Name} has an invalid bridge piece {piece}!");
+                Pieces.Add(bridgePiece);
```
Rampastring: Add an empty line after reach `throw new ...` statement in these entire `if-else` blocks, it's going to look clearer.

src/TSMapEditor/Models/Bridge.cs
```diff
@@ -0,0 +1,92 @@
...
+                string bridgeEnd = iniSection.GetStringValue($"BridgeEnd.{suffix}", null);
+                if (bridgeEnd == null)
+                    throw new BridgeLoadException($"Low bridge {bridge.Name} has no end overlay!");
+                End = rules.FindOverlayType(bridgeEnd)?.Index ??
+                      throw new BridgeLoadException($"Low bridge {bridge.Name} has an invalid end overlay {bridgeEnd}!");
+
+                Pieces = iniSection.GetListValue($"BridgePieces.{suffix}", ',', (overlayName) => rules.FindOverlayType(overlayName)?.Index ??
+                    throw new BridgeLoadException($"Low bridge {bridge.Name} has an invalid bridge piece {overlayName}!"));
+                if (Pieces.Count == 0)
+                    throw new BridgeLoadException($"Low bridge {bridge.Name} has no bridge pieces!");
```
Rampastring: Shouldn't this be also checked for high bridges? Could the check be moved outside the `if-else` structure?

src/TSMapEditor/Models/Bridge.cs
```diff
@@ -0,0 +1,92 @@
...
+                Pieces.Add(bridgePiece);
+            }
+        }
+
+        public int Start;
+        public int End;
+        public List<int> Pieces = new List<int>();
+    }
+
+    public class Bridge
```
Rampastring: I think the name of this class can be confused with a bridge that is actually placed on the map, instead of being a container for a specific type of bridge.

Renaming it to `BridgeType` would solve the issue. Kinda like we have `UnitType` and `Unit`.

src/TSMapEditor/Rendering/ObjectRenderers/OverlayRenderer.cs
```diff
@@ -25,6 +25,29 @@ protected override void Render(Overlay gameObject, int yDrawPointWithoutCellHeig
...
+                        }
+
+                        if (bridge.NorthSouth.Pieces.Contains(overlayIndex))
+                        {
+                            drawPoint.Y -= Constants.CellHeight * 2 + 1;
+                            break;
+                        }
+                    }
+                }
+            }
```
Rampastring, line 49: This is a bit ugly. Maybe overlay types should instead contain a `HighBridgeType` property that was assigned to all bridge-related overlay in the initialization phase.

That being said, this is fine for now. If it turns out to cause issues at some later point in time, we can refactor it then.

Rampastring (changes requested):
Looks good for the most part, some small things to adjust.

src/TSMapEditor/Models/Bridge.cs
ZivDero: This behavior is correct. For low bridges, there is a list of overlays that serve as sections (thus the count check), but for high bridges there is one overlay which has multiple frames instead.

```string piece = iniSection.GetStringValue($"BridgePieces.{suffix}", null);
if (piece == null)
    throw new BridgeLoadException($"High bridge {bridge.Name} has no bridge piece!");

int bridgePiece = rules.FindOverlayType(piece)?.Index ??
      throw new BridgeLoadException($"High bridge {bridge.Name} has an invalid bridge piece {piece}!");```

These checks should be sufficient.

Rampastring (approved):

PR #73, opened 2023-10-07
Added P3 Type

ZivDero:
Adds support for P3 Type used in 2 YR trigger events.

### DIFF
--- a/src/TSMapEditor/Models/TriggerCondition.cs
+++ b/src/TSMapEditor/Models/TriggerCondition.cs
@@ -26,7 +26,11 @@ namespace TSMapEditor.Models
         public string ParamToString(int index)
         {
             if (string.IsNullOrWhiteSpace(Parameters[index]))
+            {
+                if (index == MAX_PARAM_COUNT - 1)
+                    return string.Empty;
                 return "0";
+            }
 
             return Parameters[index];
         }
--- a/src/TSMapEditor/Models/TriggerCondition.cs
+++ b/src/TSMapEditor/Models/TriggerCondition.cs
@@ -40,7 +40,7 @@ namespace TSMapEditor.Models
 
         public static TriggerCondition ParseFromArray(string[] array, int startIndex, bool useP3)
         {
-            if (startIndex + DEF_PARAM_COUNT + 1 > array.Length)
+            if (startIndex + DEF_PARAM_COUNT >= array.Length)
                 return null;
 
             var triggerCondition = new TriggerCondition();
@@ -50,7 +50,7 @@ namespace TSMapEditor.Models
 
             if (useP3)
             {
-                if (startIndex + MAX_PARAM_COUNT + 1 > array.Length)
+                if (startIndex + MAX_PARAM_COUNT >= array.Length)
                     return null;
 
                 triggerCondition.Parameters[MAX_PARAM_COUNT - 1] = array[startIndex + MAX_PARAM_COUNT];
--- a/src/TSMapEditor/Models/TriggerCondition.cs
+++ b/src/TSMapEditor/Models/TriggerCondition.cs
@@ -53,7 +53,7 @@ namespace TSMapEditor.Models
                 if (startIndex + MAX_PARAM_COUNT + 1 > array.Length)
                     return null;
 
-                triggerCondition.Parameters[2] = array[startIndex + 3];
+                triggerCondition.Parameters[MAX_PARAM_COUNT - 1] = array[startIndex + MAX_PARAM_COUNT];
             }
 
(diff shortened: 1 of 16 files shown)


src/TSMapEditor/Models/TriggerCondition.cs
```diff
@@ -13,6 +13,7 @@ public class TriggerCondition : ICloneable
         public int ConditionIndex { get; set; }
         public int Parameter1 { get; set; }
         public int Parameter2 { get; set; }
+        public string Parameter3 { get; set; }
```
Rampastring: Instead of doing this, with the introduction of a third parameter, I think it'd be better to standardize the parameters into an array of strings, like on the action side.

Then re-write the handling to loop through them where necessary, and consistently use strings instead of integers.

The code duplication was already a bit bad with two parameters (my mistake for doing it that way in the first place), although tolerable. With 3 and the third one being a string, it's worse.

Rampastring (changes requested):

src/TSMapEditor/Models/TriggerCondition.cs
```diff
@@ -21,15 +38,23 @@ public TriggerCondition DoClone()
             return (TriggerCondition)MemberwiseClone();
         }
 
-        public static TriggerCondition ParseFromArray(string[] array, int startIndex)
+        public static TriggerCondition ParseFromArray(string[] array, int startIndex, bool useP3)
         {
-            if (startIndex + INI_VALUE_COUNT > array.Length)
+            if (startIndex + DEF_PARAM_COUNT + 1 > array.Length)
```
Rampastring:
```suggestion
            if (startIndex + DEF_PARAM_COUNT >= array.Length)
```

src/TSMapEditor/Models/TriggerCondition.cs
```diff
@@ -21,15 +38,23 @@ public TriggerCondition DoClone()
...
             var triggerCondition = new TriggerCondition();
             triggerCondition.ConditionIndex = Conversions.IntFromString(array[startIndex], -1);
-            triggerCondition.Parameter1 = Conversions.IntFromString(array[startIndex + 1], -1);
-            triggerCondition.Parameter2 = Conversions.IntFromString(array[startIndex + 2], -1);
+            for (int i = 0; i < DEF_PARAM_COUNT; i++)
+                triggerCondition.Parameters[i] = array[startIndex + 1 + i];
+
+            if (useP3)
+            {
+                if (startIndex + MAX_PARAM_COUNT + 1 > array.Length)
```
Rampastring:
```suggestion
                if (startIndex + MAX_PARAM_COUNT >= array.Length)
```

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -919,14 +948,27 @@ private void EventWindowDarkeningPanel_Hidden(object sender, EventArgs e)
...
+                {
+                    condition.Parameters[i] = Math.Abs((int)triggerEventType.Parameters[i].TriggerParamType).ToString(CultureInfo.InvariantCulture);
+                    continue;
+                }
 
-            if (selectEventWindow.SelectedObject.P2Type == TriggerParamType.Unused)
-                condition.Parameter2 = 0;
+                if (triggerEventType.Parameters[i].TriggerParamType == TriggerParamType.Unused)
+                {
+                    if (i == 2)
```
Rampastring: Just for some added context, I can see this magic number being confusing otherwise if one looks at this code years later (or without having knowledge on how the trigger events work).

```suggestion
                    // P3 needs to be empty instead of 0 if it's unused
                    if (i == TriggerCondition.MAX_PARAM_COUNT - 1) 
```

Rampastring (changes requested):
Good job! A few small things to improve.

src/TSMapEditor/Models/TriggerCondition.cs
```diff
@@ -8,11 +8,32 @@ namespace TSMapEditor.Models
...
+
+        public string[] Parameters { get; private set; } = new string[MAX_PARAM_COUNT];
+
+        public string ParamToString(int index)
+        {
+            if (string.IsNullOrWhiteSpace(Parameters[index]))
+            {
+                if (index == MAX_PARAM_COUNT - 1)
+                    return string.Empty;
+                return "0";
```
Rampastring, line 33: Leave a blank line between the return statements. In most cases, having a blank line after a braceless `if` looks clearer IMO.

```suggestion

                return "0";
```

Rampastring (changes requested):

Rampastring (approved):

PR #75, opened 2023-11-09
Connected Overlay Painting Tool

ZivDero:
This PR adds a tool to paint overlays that connect (i. e. walls and tracks).
The overlays to be painted are fully configurable in ConnectedOverlays.ini


src/TSMapEditor/Models/ConnectedOverlayType.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace TSMapEditor.Models
+{
+    public class ConnectedOverlayFrame
+    {
+        public OverlayType OverlayType { get; set; }
+        public int FrameIndex { get; set; }
+        public BitArray ConnectsTo { get; set; }
+    }
+    public class ConnectedOverlayType
```
Rampastring, line 19:
```suggestion

    public class ConnectedOverlayType
```

src/TSMapEditor/Models/ConnectedOverlayType.cs
```diff
@@ -0,0 +1,113 @@
...
+        public ConnectedOverlayType(IniSection iniSection, Rules rules)
+        {
+            Name = iniSection.SectionName;
+            FrameCount = iniSection.GetIntValue("Frames", 0);
+
+            if (FrameCount < 1)
+                throw new Exception($"Connected overlay type {Name} has an invalid frame count {FrameCount}!");
+
+            string connectionMaskString = iniSection.GetStringValue("ConnectionMask", null);
+            if (connectionMaskString == null || connectionMaskString.Length != 8 || Regex.IsMatch(connectionMaskString, "[^01]"))
```
Rampastring:
```suggestion
            if (connectionMaskString == null || connectionMaskString.Length != Direction.Count || Regex.IsMatch(connectionMaskString, "[^01]"))
```

src/TSMapEditor/Models/ConnectedOverlayType.cs
```diff
@@ -0,0 +1,113 @@
...
+                throw new Exception($"Connected overlay type has an invalid connection mask {connectionMaskString}!");
+            ConnectionMask = new BitArray(connectionMaskString.Select(c => c == '1').ToArray());
```
Rampastring, line 33:
```suggestion

            ConnectionMask = new BitArray(connectionMaskString.Select(c => c == '1').ToArray());
```

src/TSMapEditor/Models/ConnectedOverlayType.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            Frames = new List<ConnectedOverlayFrame>();
+
+            for (int i = 0; i < FrameCount; i++)
+            {
+                string overlayName = iniSection.GetStringValue($"Frame{i}.Overlay", null);
+                OverlayType overlayType = rules.FindOverlayType(overlayName) ??
+                              throw new Exception($"Connected overlay type {i} has an invalid overlay name {overlayName}!");
```
Rampastring: Currently this does not actually tell the type, but the frame number.

```suggestion
                              throw new Exception($"Connected overlay type {Name} has an invalid overlay type name {overlayName}!");
```

src/TSMapEditor/Config/ConnectedOverlays.ini
```diff
@@ -0,0 +1,240 @@
...
+
+[ConnectedOverlays]
+0=GDI Wall
+1=Nod Wall
+2=Sandbags
+3=Tracks
+
+; [Name]
+; ConnectionMask=BitArray(8), whether the overlay can connect in those directions at all
+; Order: NW, W, SW, N, S, NE, E, SE
```
Rampastring: Wouldn't it make more sense to follow the order of the `Direction` enum?

That would also allow looping over the `Direction` enum and using `Helpers.VisualDirectionToPoint` in the loop code in `GetOverlayForCell` instead of writing a nested `for` loop.

src/TSMapEditor/Mutations/Classes/PlaceConnectedOverlayMutation.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Collections.Generic;
+using TSMapEditor.GameMath;
+using TSMapEditor.Models;
+using TSMapEditor.Rendering;
+using TSMapEditor.UI;
+
+namespace TSMapEditor.Mutations.Classes
+{
+    /// <summary>
+    /// A mutation that allows placing regular, individual overlay.
```
Rampastring: Copy/paste mistake in comment

src/TSMapEditor/Mutations/Classes/PlaceConnectedOverlayMutation.cs
```diff
@@ -0,0 +1,128 @@
...
+                        var originalTile = MutationTarget.Map.GetTile(cellCoords + new Point2D(xOffset, yOffset));
+                        if (originalTile == null)
+                            continue;
+
+                        originalOverlayInfos.Add(new OriginalOverlayInfo()
+                        {
+                            CellCoords = originalTile.CoordsToPoint(),
+                            OverlayTypeIndex = originalTile.Overlay?.OverlayType.Index ?? -1,
+                            FrameIndex = originalTile.Overlay?.FrameIndex ?? -1,
+                        });
```
Rampastring: Won't this potentially add the original overlay info for the same cells multiple times if BrushSize is bigger than 1x1?

As far as I can see, this entire nested `for` loop could be avoided by using `DoForBrushSizeAndSurroundings` and its offset. Possibly with two calls to the function, one for adding the undo data and another for setting the connected overlay.

Rampastring (changes requested):
Some stylistical and some implementation-related things to improve.

src/TSMapEditor/Models/ConnectedOverlayType.cs
```diff
@@ -36,15 +38,16 @@ public ConnectedOverlayType(IniSection iniSection, Rules rules)
...
                 int frameIndex = iniSection.GetIntValue($"Frame{i}.FrameIndex", -1);
                 if (frameIndex < 0)
-                    throw new Exception($"Connected overlay type {i} has an invalid frame index {frameIndex}!");
+                    throw new Exception($"Connected overlay type {Name}, frame {i} has an invalid frame index {frameIndex}!");
 
                 string connectsToString = iniSection.GetStringValue($"Frame{i}.ConnectsTo", null);
-                if (connectsToString == null || connectsToString.Length != 8 || Regex.IsMatch(connectsToString, "[^01]"))
-                    throw new Exception($"Connected overlay type {i} has an invalid ConnectsTo mask {connectsToString}!");
+                if (connectsToString == null || connectsToString.Length != (int)Direction.Count || Regex.IsMatch(connectsToString, "[^01]"))
+                    throw new Exception($"Connected overlay type {Name}, frame {i} has an invalid ConnectsTo mask {connectsToString}!");
```
Rampastring: Throw `INIConfigException` instead of the generic exception type. Applies to all the exceptions thrown here.

Rampastring (changes requested):
Very nice and clean implementation, one nitpick about exception types left.

src/TSMapEditor/Models/ConnectedOverlayType.cs
ZivDero: Done

Rampastring (approved):

PR #80, opened 2023-12-27
Allowed connected overlays to connect to other connected overlays

ZivDero:
Adds "related overlays" that a connected overlay can also connect to.
NOTE: The config changes are for TS-Client, so upon merging they should be reverted and reapplied to TS-Client.


src/TSMapEditor/Config/ConnectedOverlays.ini
```diff
@@ -25,6 +24,7 @@
 
 [Concrete Wall]
 ConnectionMask=10101010
+RelatedOverlays=GDI Wall (Damaged);,GDI Wall (Very Damaged)
```
Rampastring: Typing entire UI names here can get a bit error-prone I think.

It would be a good idea to introduce a `UIName` property to connected overlay types, read it from these sections, and display its value in the editor's sidebar (with a fallback to regular `Name=` if not defined), so you could have separate INI and UI names for connected overlays.

Then you could define for example `0=GAWALL`, `1=GAWALL_DAMAGED`, and have `UIName=Concrete Wall` under the `[GAWALL]` section, and `RelatedOverlays=GAWALL_DAMAGED` etc.

Rampastring (commented):
Looks good, there's one suggestion.

Rampastring (approved):

PR #81, opened 2023-12-27
Fix: Make inheritance fully recursive

MortonPL:
Current implementation of Ares/Phobos INI section inheritance handles only one level of inheritance. This PR makes inheritance procedure fully recursive, allowing several layers of inheritance ([GRANDPARENT]->[PARENT]->[CHILD] etc.).


src/TSMapEditor/Extensions/IniFileEx.cs
```diff
@@ -14,6 +16,8 @@ public class IniFileEx: IniFile
     private static readonly string PhobosIncludeSection = "$Include";
     private static readonly string PhobosInheritsSection = "$Inherits";
 
+    private List<IniSectionEx> ExSections = new();
```
Rampastring: Couldn't a `Dictionary<IniSection, bool>` be used to achieve the same, but with better performance and without the need for an extended class?

MortonPL: You're right, I overcomplicated it. I would use a `HashSet` instead of a `Dictionary`, though.

Rampastring (approved):

src/TSMapEditor/Extensions/IniFileEx.cs
Rampastring: Good point.

PR #96, opened 2024-01-10
Stringtable parsing and StringTableEntry trigger action param

MortonPL:
* Add `CsfFile` that can read RA2/YR CSF stringtable files,
* Add `[Stringtables]` section to `Config/FileManagerConfig.ini` that specifies which CSF files have to be loaded on startup,
* Add `StringTableParam` trigger action type param type with `SelectStringWindow` GUI.

YR configuration provided in PR #105.

image

### DIFF
--- a/src/TSMapEditor/CCEngine/CCFileManager.cs
+++ b/src/TSMapEditor/CCEngine/CCFileManager.cs
@@ -35,7 +35,7 @@ namespace TSMapEditor.CCEngine
             iniFile.DoForEveryValueInSection("SearchDirectories", v => AddSearchDirectory(Path.Combine(GameDirectory, v)));
             iniFile.DoForEveryValueInSection("PrimaryMIXFiles", v => LoadPrimaryMixFile(v));
             iniFile.DoForEveryValueInSection("SecondaryMIXFiles", v => LoadSecondaryMixFile(v));
-            iniFile.DoForEveryValueInSection("Stringtables", v => LoadStringtable(v));
+            iniFile.DoForEveryValueInSection("StringTables", v => LoadStringTable(v));
         }
 
         /// <summary>
@@ -154,7 +154,7 @@ namespace TSMapEditor.CCEngine
         /// Throws a FileNotFoundException if the CSF file isn't found.
         /// </summary>
         /// <param name="name">The name of the CSf file.</param>
-        public void LoadStringtable(string name)
+        public void LoadStringTable(string name)
         {
             var data = LoadFile(name);
             if (data == null)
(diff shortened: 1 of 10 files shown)


src/TSMapEditor/CCEngine/CsfFile.cs
```diff
@@ -0,0 +1,232 @@
...
+            if (!skip)
+            {
+                var stringBuffer = new byte[stringLength * 2];
+                memoryStream.Read(stringBuffer, 0, stringBuffer.Length);
+                // Westwood's "encoding".
+                for (uint i = 0; i < stringBuffer.Length; i++)
+                    stringBuffer[i] = (byte)~stringBuffer[i];
+
+                csfString = System.Text.Encoding.Unicode.GetString(stringBuffer);
+            } else
```
Rampastring:
```suggestion
            }
            else
```

src/TSMapEditor/Models/Map.cs
```diff
@@ -171,6 +171,8 @@ public Rectangle LocalSize
         public ITheater TheaterInstance { get; set; }
         public string LoadedTheaterName => TheaterInstance.Theater.UIName;
 
+        public Dictionary<string, CsfString> StringTable { get; private set; } = new();
```
Rampastring: Since you can't insert new CSF strings into a map and this class is mostly for holding and handling _map_ data, it would make more sense to either move this and the related functionality to Rules or create another class/object for holding the CSF lines.

src/TSMapEditor/CCEngine/CCFileManager.cs
```diff
@@ -143,6 +149,31 @@ public void LoadSecondaryMixFile(string name)
...
+
+        /// <summary>
+        /// Drain loaded stringtable list and return it.
+        /// </summary>
+        public List<CsfFile> DrainStringTables()
+        {
+            var list = csfFiles;
+            csfFiles = new();
+            return list;
+        }
```
Rampastring, line 165: What is the point of clearing the stringtable list here?

Rampastring (changes requested):

src/TSMapEditor/Models/Map.cs
MortonPL: Moved to `TSMapEditor.Models.StringTable`, with Map having a public StringTable property.

src/TSMapEditor/CCEngine/CCFileManager.cs
MortonPL: No point besides moving the ownership from CCFileManager to Map. Removed.

Rampastring (approved):

PR #104, opened 2024-01-16
Improvements to building animations

ZivDero:
- Add support for IdleAnims. This makes buildings like ConYards, naval yard, the grinder, etc. render in full
- Add support for SuperAnims (only the first one, which represents the "not ready" state). Previously, some supers got rendered partially, and some, like the YR Nuke, not at all, with all their graphics contained in a SuperAnim.
- Currently suffers from issues with Y-Sorting

### DIFF
--- a/src/TSMapEditor/Models/ArtConfig/BuildingArtConfig.cs
+++ b/src/TSMapEditor/Models/ArtConfig/BuildingArtConfig.cs
@@ -196,6 +196,13 @@ namespace TSMapEditor.Models.ArtConfig
         }
     }
 
+    public struct BuildingAnimType
+    {
+        public string ININame { get; set; }
+        public int YSort { get; set; }
+        public int ZAdjust { get; set; }
+    }
+
     public class BuildingArtConfig : IArtConfig
     {
         public BuildingArtConfig() { }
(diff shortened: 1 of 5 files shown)


src/TSMapEditor/Models/ArtConfig/BuildingArtConfig.cs
```diff
@@ -231,16 +231,33 @@ public void ReadFromIniSection(IniSection iniSection)
...
+
+            var buildingAnimClasses = new Dictionary<string, string[]>
             {
-                string animTypeName = iniSection.GetStringValue("ActiveAnim" + i, null);
-                if (string.IsNullOrEmpty(animTypeName))
-                    break;
+                { "ActiveAnim", new [] { "", "Two", "Three", "Four" } },
+                { "IdleAnim", new [] { "", "Two" } },
+                { "SuperAnim", new [] { "" } }
+            };
```
Rampastring: This dictionary should be made a static member of the class as it's always identical.

src/TSMapEditor/Models/ArtConfig/BuildingArtConfig.cs
```diff
@@ -208,7 +208,7 @@ public class BuildingArtConfig : IArtConfig
         public bool Theater { get; set; }
         public string Image { get; set; }
         public string BibShape { get; set; }
-        public string[] AnimNames { get; set; } = Array.Empty<string>();
+        public Dictionary<string, (int YSort, int ZAdjust)> AnimNames { get; set; } = new();
```
Rampastring: Not a fan of calling this `AnimNames` as it's not only names anymore. 

Also, I don't think this should be a dictionary as it's never accessed per-key, like dictionaries typically are. Rather make a `struct` for containing the anim properties and make an array of the struct.

Rampastring (changes requested):
A couple of changes required.

Rampastring (approved):

PR #107, opened 2024-01-19
"Count All Credits On Map" C# script and Tiberium/ore bug fixes

### LINKED ISSUE #98: Count ore/tiberium value globally
The current tool for counting credits cannot be used map-wide, because the map is a rectangle and the tool selection area is a a rotated rectangle (diamond). FS/FA2 always display credits on map, here it could be implemented as a C# script or a Tools menu entry.

MortonPL:
Closes #98 by introducing a new C# script, "Count All Credits On Map".

image

Fixes an issue where Tiberium type 1 and 2 are swapped for YR.

Fixes an issue where the "Calculate Credits..." cursor action doesn't count the last row and column of selection.

### DIFF
--- a/src/TSMapEditor/Config/Scripts/Count Credits On Map.cs
+++ b/src/TSMapEditor/Config/Scripts/Count Credits On Map.cs
@@ -44,7 +44,7 @@ namespace WAEScript
                 if (cell.Overlay == null || cell.Overlay.OverlayType == null || !cell.Overlay.OverlayType.Tiberium)
                     return;
 
-                int tiberiumIndex = cell.Overlay.OverlayType.GetTiberiumIndex();
+                int tiberiumIndex = cell.Overlay.OverlayType.GetTiberiumIndex(Constants.UseCountries);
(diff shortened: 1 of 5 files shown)


src/TSMapEditor/Config/Scripts/Count Credits On Map.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// The function that actually does the magic.
+        /// </summary>
+        /// <param name="map">Map argument that allows us to access map data.</param>
+        public void Perform(Map map)
+        {
+            count = 0;
+
+            map.DoForAllValidTiles(cell =>
+            {
+                if (cell.Overlay == null || cell.Overlay.OverlayType == null || !cell.Overlay.OverlayType.Tiberium)
```
ZivDero: ```cell.Overlay?.OverlayType == null```

ZivDero (approved):

Rampastring (approved):

PR #109, opened 2024-01-20
Replace naval unit category check

ZivDero:
The current check is overly unstraightforward and leads to miscategorization in YR. The proposed check is meant to be as functional for both TS and YR.


Rampastring (changes requested):
This wouldn't cut it for DTA which has hover and amphibious units with `MovementZone=Fly` due to TS engine limitations.

PR #110, opened 2024-01-20
Improve building animation y-sorting and enable damaged frames

ZivDero:
- Make building animations sort with the same "score" as their buildings, and when compared against their buildings use the extra sorting modifiers;
- Make building animations display damaged frames when their building is low on HP.

### DIFF
--- a/src/TSMapEditor/UI/Sidebar/OverlayListPanel.cs
+++ b/src/TSMapEditor/UI/Sidebar/OverlayListPanel.cs
@@ -169,7 +169,7 @@ namespace TSMapEditor.UI.Sidebar
                         int frameCount = textures.GetFrameCount();
                         int frameNumber = firstEntry.Frame;
                         if (firstEntry.OverlayType.Tiberium)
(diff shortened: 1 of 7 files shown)


src/TSMapEditor/UI/Sidebar/OverlayListPanel.cs
```diff
@@ -169,7 +169,7 @@ private void InitOverlays()
                         int frameCount = textures.GetFrameCount();
                         int frameNumber = firstEntry.Frame;
                         if (firstEntry.OverlayType.Tiberium)
-                            frameNumber = (frameCount / 2) - 1;
+                            frameNumber = frameCount / 2 - 1;
```
Rampastring: I prefer having parentheses here to make the intended calculation order clearer, especially for potential contributors who don't have a strong background in math.

ZivDero: Restored the parentheses

Rampastring (approved):

Issue #88, opened 2024-01-08
Don't register duplicate TechnoTypes


ZivDero:
If a TechnoType is listed in rules twice, it shouldn't be registered twice, as the game doesn't do that either.
Exists in vanilla YR with NAPSYA.

![image](https://github.com/Rampastring/TSMapEditor/assets/5639388/89778e1e-4447-475d-af36-6d8a3c5469f5)

PR #113, opened 2024-01-21
Skip adding duplicate types

ZivDero:
- Add a check to skip adding types if they are being redefined in the rules (occurs once in vanilla YR)
Addresses #88

### DIFF
--- a/src/TSMapEditor/Models/Rules.cs
+++ b/src/TSMapEditor/Models/Rules.cs
@@ -313,6 +313,11 @@ namespace TSMapEditor.Models
 
                 var objectType = typeof(T);
 
+                // If we encounter the same object listed twice, don't create a duplicate
+                var iniNameProperty = objectType.GetProperty("ININame");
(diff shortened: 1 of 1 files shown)


src/TSMapEditor/Models/Rules.cs
```diff
@@ -313,6 +313,11 @@ private void InitFromTypeSection<T>(IniFile iniFile, string sectionName, List<T>
 
                 var objectType = typeof(T);
 
+                // If we encounter the same object listed twice, don't create a duplicate
+                var iniNameProperty = objectType.GetProperty("ININame");
+                if (targetList.Find(t => iniNameProperty?.GetValue(t)?.ToString() == typeName) != null)
```
Rampastring: Use `targetList.Exists` instead.

Rampastring (changes requested):

Rampastring (approved):

PR #115, opened 2024-01-21
Use string interpolation where possible

ZivDero:
This PR replaces all instances of string concatenation (and string.Format) with string interpolation.
- String interpolation is marginally more efficient as it uses StringBuilder
- Interpolated strings are easier to read

### DIFF
--- a/src/TSMapEditor/CCEngine/ShpFile.cs
+++ b/src/TSMapEditor/CCEngine/ShpFile.cs
@@ -29,7 +29,7 @@ namespace TSMapEditor.CCEngine
         public ShpFileHeader(byte[] buffer)
         {
             if (buffer.Length < SizeOf)
-                throw new ShpLoadException(nameof(ShpFileHeader) + ": buffer is not long enough");
+                throw new ShpLoadException($"{nameof(ShpFileHeader)}: buffer is not long enough");
 
             Unknown = BitConverter.ToUInt16(buffer, 0);
             if (Unknown != 0)
@@ -56,7 +56,7 @@ namespace TSMapEditor.CCEngine
         public ShpFrameInfo(Stream stream)
         {
             if (stream.Length < stream.Position + SizeOf)
-                throw new ShpLoadException(nameof(ShpFrameInfo) + ": buffer is not long enough");
+                throw new ShpLoadException($"{nameof(ShpFrameInfo)}: buffer is not long enough");
 
             XOffset = ReadUShortFromStream(stream);
             YOffset = ReadUShortFromStream(stream);
@@ -158,7 +158,8 @@ namespace TSMapEditor.CCEngine
             }
             catch (ShpLoadException ex)
             {
-                throw new ShpLoadException("Failed to load SHP file. Make sure that the file is not corrupted. Filename: " + fileName + ", original exception: " + ex.Message);
+                throw new ShpLoadException(
+                    $"Failed to load SHP file. Make sure that the file is not corrupted. Filename: {fileName}, original exception: {ex.Message}");
             }
         }
 
@@ -204,8 +205,8 @@ namespace TSMapEditor.CCEngine
 
                 if (lineDataStartOffset + lineDataLength > fileData.Length || lineDataLength < 2)
                 {
-                    throw new ShpLoadException($"Line data length out-of-bounds in SHP RLE-Zero frame. " +
-                        $"File name: {fileName}, frame index: {frameIndex}, line data length: {lineDataLength}");
+                    throw new ShpLoadException(
+                        $"Line data length out-of-bounds in SHP RLE-Zero frame. File name: {fileName}, frame index: {frameIndex}, line data length: {lineDataLength}");
                 }
 
                 // Line length includes the two-byte line data length value, skip it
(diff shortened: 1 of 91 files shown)


src/TSMapEditor/CCEngine/VxlFile.cs
```diff
@@ -44,11 +44,12 @@ public VxlFile(byte[] buffer, string filename = "") : base(new MemoryStream(buff
...
         {
             if (Length < FileHeader.Size)
-                throw new VxlLoadException(nameof(VxlFile) + " .vxl is shorter than specified in its header!");
+                throw new VxlLoadException($"{nameof(VxlFile)} .vxl is shorter than specified in its header!");
 
             Header.Read(this);
             if (Header.HeaderCount == 0 || Header.TailerCount == 0 || Header.TailerCount != Header.HeaderCount)
-                throw new VxlLoadException(nameof(VxlFile) + " .vxl has no header or tailer sections, or their count doesn't match!");
+                throw new VxlLoadException(
+                    $"{nameof(VxlFile)} .vxl has no header or tailer sections, or their count doesn't match!");
```
Rampastring, line 52: Splitting this into another line seems pointless. It also breaks our contribution guidelines.

ZivDero:
Looking back, I will go over this in detail and restore concatenation in some places where it makes it more readable (mostly very long lines with breaks).

src/TSMapEditor/CCEngine/ShpFile.cs
```diff
@@ -254,8 +255,8 @@ private void DecompressRLEZero(byte[] frameData, int frameIndex, ShpFrameInfo fr
                     {
                         if (pixelPositionOnLine >= frameInfo.Width)
                         {
-                            throw new ShpLoadException($"Out-of-bounds pixel position on color data in SHP RLE-Zero frame. " +
-                                $"File name: {fileName}, frame index: {frameIndex}, line index: {lineIndex}, file offset: {lineDataStartOffset + currentByteIndex}");
+                            throw new ShpLoadException(
+                                $"Out-of-bounds pixel position on color data in SHP RLE-Zero frame. File name: {fileName}, frame index: {frameIndex}, line index: {lineIndex}, file offset: {lineDataStartOffset + currentByteIndex}");
```
Rampastring, line 259: I think the original formatting here was superior with it not having such a long line.

src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -84,8 +84,8 @@ public static void PostCheckMap(IMap map, TheaterGraphics theaterGraphics)
                 int maxSubTileIndex = tile.SubTileCount - 1;
                 if (t.SubTileIndex > maxSubTileIndex)
                 {
-                    AddMapLoadError($"Invalid sub-tile index {t.SubTileIndex} for cell at {t.CoordsToPoint()} (max: {maxSubTileIndex}) - setting it to 0. " +
-                        $"TileSet: {tileSet.SetName} ({tileSet.FileName}), index of tile within its set: {tile.TileIndexInTileSet}");
+                    AddMapLoadError(
+                        $"Invalid sub-tile index {t.SubTileIndex} for cell at {t.CoordsToPoint()} (max: {maxSubTileIndex}) - setting it to 0. TileSet: {tileSet.SetName} ({tileSet.FileName}), index of tile within its set: {tile.TileIndexInTileSet}");
```
Rampastring, line 88: I think the original formatting here was superior with it not having such a long line.

src/TSMapEditor/Initialization/MapWriter.cs
```diff
@@ -599,17 +586,8 @@ public static void WriteBuildings(IMap map, IniFile mapIni)
...
-                               $"{BoolToObjectStyle(structure.AISellable)}," +
-                               $"{BoolToObjectStyle(structure.AIRebuildable)}," +
-                               $"{BoolToObjectStyle(structure.Powered)}," +
-                               $"{structure.UpgradeCount}," +
-                               $"{(int)structure.Spotlight}," + 
-                               $"{upgrade1},{upgrade2},{upgrade3}," +
-                               $"{BoolToObjectStyle(structure.AIRepairable)}," +
-                               $"{BoolToObjectStyle(structure.Nominal)}";
+                string value =
+                    $"{structure.Owner.ININame},{structure.ObjectType.ININame},{structure.HP},{structure.Position.X},{structure.Position.Y},{structure.Facing},{attachedTag},{BoolToObjectStyle(structure.AISellable)},{BoolToObjectStyle(structure.AIRebuildable)},{BoolToObjectStyle(structure.Powered)},{structure.UpgradeCount},{(int)structure.Spotlight},{upgrade1},{upgrade2},{upgrade3},{BoolToObjectStyle(structure.AIRepairable)},{BoolToObjectStyle(structure.Nominal)}";
```
Rampastring, line 590: Ew.

src/TSMapEditor/Rendering/GameClass.cs
```diff
@@ -50,29 +51,28 @@ public GameClass()
...
             }
 
             Logger.Log("Exiting.");
 
             windowManager?.HideWindow();
-            System.Windows.Forms.MessageBox.Show("The map editor has crashed." + Environment.NewLine + Environment.NewLine +
-                "Exception information logged into except.txt:" + Environment.NewLine + Environment.NewLine +
-                sb.ToString());
+            System.Windows.Forms.MessageBox.Show(
+                $"The map editor has crashed.{Environment.NewLine}{Environment.NewLine}Exception information logged into except.txt:{Environment.NewLine}{Environment.NewLine}{sb}");
```
Rampastring, line 75: The original formatting here nicely split the lines just like they are split in the resulting message box.

Rampastring (commented):
I actually disagree about interpolated strings being easier to read in all cases. For example

```cs
throw new FileNotFoundException("Primary MIX file not found: " + name);
throw new FileNotFoundException($"Primary MIX file not found: {name}");
```

In cases like this there's no clear difference. I actually find the non-interpolated slightly easier to read, although likely only because I'm used to it.

Where interpolation is clearly superior is when there's multiple arguments to the string. Such as this case

```cs
$"Failed to load CSF file. Make sure that the file is not corrupted. Filename: {fileName}, original exception: {ex.Message}"
```

This means that interpolation itself can be used in all cases without any real drawbacks. Which means that I could approve the core idea of this PR.

However, outside of interpolation itself, this PR has some very ugly newline changes, probably as a result of doing this change with some tool that uses "generic" code formatting rules without caring about the context. I highlighted a few, but there are far more in the code.

This is why I'm generally careful with code refactoring tools and don't like code style enforcers, as they tend to always do something worse compared to my own, manual work.

PR #119, opened 2024-02-03
Add voxel turret and barrel rendering

ZivDero:
This PR adds the support to render voxel turrets and barrels, making voxels support mostly complete.


Rampastring (approved):
VRAM consumption can increase significantly due to the bounds hack, which is a bit concerning. Aside from that it looks good.

PR #123, opened 2024-02-18
Make building animations be rendered by the building renderer

ZivDero:
- Make building animations be rendered by the building renderer (using a sub-renderer)
- Fix sorting issues with building animations (z-sorting is a guess so the multiplier may need to be adjusted, but probably not)
- Make component tower upgrades be displayed

### DIFF
--- a/src/TSMapEditor/Models/ArtConfig/BuildingArtConfig.cs
+++ b/src/TSMapEditor/Models/ArtConfig/BuildingArtConfig.cs
@@ -244,7 +244,7 @@ namespace TSMapEditor.Models.ArtConfig
         public bool Theater { get; set; }
         public string Image { get; set; }
         public string BibShape { get; set; }
-        public List<BuildingAnimArtConfig> BuildingAnimsConfigs { get; set; } = new();
+        public List<BuildingAnimArtConfig> BuildingAnimConfigs { get; set; } = new();
         public List<PowerUpAnimArtConfig> PowerUpAnimConfigs { get; set; } = new();
         public AnimType[] Anims { get; set; } = Array.Empty<AnimType>();
         public AnimType[] PowerUpAnims { get; set; } = Array.Empty<AnimType>();
@@ -292,7 +292,7 @@ namespace TSMapEditor.Models.ArtConfig
                 }
             }
 
-            BuildingAnimsConfigs = anims;
+            BuildingAnimConfigs = anims;
         }
 
         public void ReadUpgradeAnims(int upgradeCount, IniSection iniSection)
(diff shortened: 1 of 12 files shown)


src/TSMapEditor/Models/ArtConfig/BuildingArtConfig.cs
```diff
@@ -233,21 +244,22 @@ public class BuildingArtConfig : IArtConfig
         public bool Theater { get; set; }
         public string Image { get; set; }
         public string BibShape { get; set; }
-        public List<BuildingAnimType> BuildingAnimTypes { get; set; } = new();
-        public List<PowerUpAnimConfig> PowerUpAnims { get; set; } = new();
+        public List<BuildingAnimArtConfig> BuildingAnimsConfigs { get; set; } = new();
```
Rampastring:
```suggestion
        public List<BuildingAnimArtConfig> BuildingAnimConfigs { get; set; } = new();
```

src/TSMapEditor/Models/Rules.cs
```diff
@@ -419,20 +420,38 @@ private void SolveBuildingTypeDependencies(BuildingType type)
...
+            var powerUpAnims = new List<AnimType>();
+
+            foreach (var powerUpType in BuildingTypes.Where(bt => !string.IsNullOrWhiteSpace(bt.PowersUpBuilding) &&
+                                                                  bt.PowersUpBuilding.Equals(type.ININame, StringComparison.OrdinalIgnoreCase)))
+            {
+                string image = string.IsNullOrWhiteSpace(powerUpType.ArtConfig.Image) ?
+                               string.IsNullOrWhiteSpace(powerUpType.Image) ?
+                               powerUpType.ININame :
+                               powerUpType.Image :
+                               powerUpType.ArtConfig.Image;
```
Rampastring: I have always disliked nested ternary operators. Please rewrite this with `if-else` conditions.

Yes, ternaries are shorter but if there's multiple, they get much harder to parse compared to `if-else` to the point the reduction in syntax isn't worth it anymore.

src/TSMapEditor/Rendering/ObjectRenderers/AnimRenderer.cs
```diff
@@ -36,8 +36,9 @@ protected override void Render(Animation gameObject, int heightOffset, Point2D d
             if (gameObject.IsTurretAnim)
             {
                 // Turret anims have their facing frames reversed
+                // Turret anims also only have 32 facings. GACTWR_C is broken, so we have to hardcode that
```
Rampastring: GACTWR_C is also broken in-game (in vanilla TS, not TSC). I don't think it's relevant to mention here. As long as WAE matches in-game - even if in-game is broken - it's good.

Rampastring (changes requested):
Looks mostly good, a couple small comments.

Rampastring (approved):

PR #130, opened 2024-03-03
Fixes for minor issues with YR/Ares/Phobos

Starkku:
- Do not draw palette color #0 on voxels
- Fix empty voxel sections potentially causing an attempt to create zero-width/height `RenderTarget` texture
- Render voxel turret even if main voxel body is missing
- Render buildings even if their main shape file is empty or missing
- Improved palette selection for objects for YR/Ares/Phobos (supports `Palette` for all technos and wall OverlayTypes)
- Correctly use building's palette for building anims unless `ShouldUseCellDrawer` is set to false (this affects TS as well, as it works identically there)

### DIFF
--- a/src/TSMapEditor/Models/ArtConfig/AnimArtConfig.cs
+++ b/src/TSMapEditor/Models/ArtConfig/AnimArtConfig.cs
@@ -14,10 +14,15 @@ namespace TSMapEditor.Models.ArtConfig
         public bool Theater { get; set; }
         public bool AltPalette { get; set; }
         public string CustomPalette { get; set; } // Ares
-        public bool ShouldUseCellDrawer { get; set; } = true; // Only used by building and tile animations
         public bool Shadow { get; set; }
         public int Start { get; set; }
         public int Translucency { get; set; }
+
+        /// <summary>
+        /// Only used on building and tile animations, setting it to false makes them draw
+        /// using regular animation palette (or a custom animation palette if available).
+        /// </summary>
+        public bool ShouldUseCellDrawer { get; set; } = true;
 
         /// <summary>
         /// Not an INI entry. Temporarily set per-type instead of per instance until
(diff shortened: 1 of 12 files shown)


Starkku:
>- Render voxel turret even if main voxel body is missing
>- Render buildings even if their main shape file is empty or missing

Realized that these can (and might actually be better off, because of how the sidebar preview textures are fetched currently) fixed via rules/art overrides. Consequently, I could drop them off this PR if needed.

src/TSMapEditor/Rendering/ObjectRenderers/ObjectRenderer.cs
```diff
@@ -58,7 +58,13 @@ public void Draw(T gameObject, bool checkInCamera)
...
                 return;
 
-            if (frame == null && ShouldRenderReplacementText(gameObject))
+            bool forceRender = false;
+
+            // Draw buildings that do not have main shape set.
+            if (gameObject as Structure is var structure && structure != null)
+                forceRender = structure.Anims.Length > 0 || structure.TurretAnim != null;
+
+            if ((frame == null && !forceRender) && ShouldRenderReplacementText(gameObject))
```
Rampastring: Is there any good reason why you didn't modify the override of `ShouldRenderReplacementText` in `BuildingRenderer` and instead added this hack?

src/TSMapEditor/Models/ArtConfig/AnimArtConfig.cs
```diff
@@ -14,6 +14,7 @@ public class AnimArtConfig : IArtConfig
         public bool Theater { get; set; }
         public bool AltPalette { get; set; }
         public string CustomPalette { get; set; } // Ares
+        public bool ShouldUseCellDrawer { get; set; } = true; // Only used by building and tile animations
```
Rampastring: Make this comment a `/// <summary>` style comment instead so it is easier to check from references.

Rampastring (changes requested):
Looks good in general, a couple things to fix or explain.

Also, this PR breaks the contribution guidelines on PR scope, but with only 143 modified lines it's not too bad, so I'll let it pass this time.

src/TSMapEditor/Rendering/ObjectRenderers/ObjectRenderer.cs
Starkku: No, so I will change it.

Rampastring (approved):

PR #136, opened 2024-03-13
Improvements to the YR config

ZivDero:
- Fix the unit palettes for different theaters (all point to unittem.pal right now).
- Clean up the extra ini file string in Contants.ini (optional), looks nicer IMO.
- Fix script actions DoThis to be accurate with YR #142


Rampastring:
Applied these changes aside from the Constants.ini one. I think for situations of troubleshooting and for reasons of general knowledge of WAE's inner logic, it's good for YR modders using WAE to also realize that it uses the Firestorm rules addition system.

PR #138, opened 2024-03-14
Add Chinese document

MapleEve:


Rampastring:
Hey,

I'm open to having a Chinese readme, however I think it should be placed in the `docs` folder now that one exists.

SadPencil:
I believe it could be inappropriate to use `泰伯利亚黎明` as the Chinese translation of DTA because it has been associated with "Tiberium Dawn" for over a decade.

Moreover, the entire Chinese document reads as if it were translated by software, if not worse.

Rampastring:
@SadPencil That's surprising, I expected MapleEve to have spent more effort to it considering he's also Chinese. Oh well, thanks for the review.

SadPencil:
@MapleEve Hi, thank you for your effort. However, I recommend a complete rewrite as there are numerous issues in the current version that are hard to list. Please do your best to make the Chinese translation smooth and fluent. You can find a good example of Chinese documentation here: <https://github.com/XiaoMi/ha_xiaomi_home/blob/b955c199fca824922453090dad3099bdd47fc9b0/doc/README_zh.md>.

PR #140, opened 2024-03-14
Allow setting custom Event, Action and Script Action IDs

ZivDero:
Allow setting custom Event, Action and Script Action IDs required for Phobos.
Usage:
```
[Event/Action/ScriptAction]
IDOverride=500
```

### DIFF
--- a/src/TSMapEditor/Models/EditorConfig.cs
+++ b/src/TSMapEditor/Models/EditorConfig.cs
@@ -169,8 +169,10 @@ namespace TSMapEditor.Models
                 scriptAction.ReadIniSection(scriptSection);
 
                 if (ScriptActions.ContainsKey(scriptAction.ID))
-                    throw new INIConfigException(
-                        $"Error while adding Script Action {scriptAction.Name}: a Script Action with ID {scriptAction.ID} already exists!");
+                {
+                    throw new INIConfigException($"Error while adding Script Action {scriptAction.Name}: " + 
+                                                 $"a Script Action with ID {scriptAction.ID} already exists!");
+                }
 
                 ScriptActions.Add(scriptAction.ID, scriptAction);
             }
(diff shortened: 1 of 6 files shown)


src/TSMapEditor/Models/EditorConfig.cs
```diff
@@ -168,7 +168,11 @@ private void ReadScriptActions()
                 var scriptSection = iniFile.GetSection(sections[i]);
                 scriptAction.ReadIniSection(scriptSection);
 
-                ScriptActions.Add(scriptAction.Index, scriptAction);
+                if (ScriptActions.ContainsKey(scriptAction.ID))
+                    throw new INIConfigException(
+                        $"Error while adding Script Action {scriptAction.Name}: a Script Action with ID {scriptAction.ID} already exists!");
```
Rampastring:
```suggestion
                if (ScriptActions.ContainsKey(scriptAction.ID))
                {
                    throw new INIConfigException($"Error while adding Script Action {scriptAction.Name}: " + 
                        $"a Script Action with ID {scriptAction.ID} already exists!");
                }
```

As stated in the contribution guide

> Braceless code block bodies should be made only when both code block head and body are single line. Statements that split into multiple lines and nested braceless blocks are not allowed within braceless blocks

Also I think splitting the string just looks easier to read, since the string itself is so long.

src/TSMapEditor/Models/EditorConfig.cs
```diff
@@ -183,6 +187,10 @@ private void ReadTriggerEventTypes()
                 var section = iniFile.GetSection(sections[i]);
                 triggerEventType.ReadPropertiesFromIniSection(section);
 
+                if (TriggerEventTypes.ContainsKey(triggerEventType.ID))
+                    throw new INIConfigException(
+                        $"Error while adding Trigger Event {triggerEventType.Name}: a Trigger Event with ID {triggerEventType.ID} already exists!");
```
Rampastring: See style note above

src/TSMapEditor/Models/EditorConfig.cs
```diff
@@ -198,6 +206,10 @@ private void ReadTriggerActionTypes()
                 var section = iniFile.GetSection(sections[i]);
                 triggerActionType.ReadPropertiesFromIniSection(section);
 
+                if (TriggerActionTypes.ContainsKey(triggerActionType.ID))
+                    throw new INIConfigException(
+                        $"Error while adding Trigger Action {triggerActionType.Name}: a Trigger Action with ID {triggerActionType.ID} already exists!");
```
Rampastring: See style note above

Rampastring (changes requested):
A few stylistic things to fix.

Rampastring (approved):

PR #146, opened 2024-04-04
Allow specific tilesets to connect to specific LAT types in editor

Starkku:
Allows LAT grounds to treat certain arbitrary tilesets as something that they can connect to (e.g not use transition tiles in between). Tileset indices can be listed for custom LAT grounds via `GroundNConnectTo` key and for the base 4 LAT types via keys (`Rough/Sand/Pave/GreenConnectToTileSets`) read from each theater in `Theaters.ini`. Latter is because these might be something that people want to change without touching the actual theater configuration INI game uses.

This is mostly something that can be used to fix sand & pavement LAT interactions in YR but I imagine it can potentially be used to great effect with the `GroundN` stuff elsewhere.

To apply the fixes for sand<->shores & pavement<->road & misc. pave tiles in YR config to match game's own behaviour with these tiles, set following for all theaters except LUNAR (and maybe URBAN although in this case it shouldn't have adverse effects either)
```ini
PaveConnectToTileSets=20,38,40
GreenConnectToTileSets=12
```

### DIFF
--- a/src/TSMapEditor/CCEngine/Theater.cs
+++ b/src/TSMapEditor/CCEngine/Theater.cs
@@ -5,6 +5,7 @@ using System.IO;
 using TSMapEditor.Models;
 using TSMapEditor.UI;
 using TSMapEditor.Extensions;
+using System.Linq;
 
 namespace TSMapEditor.CCEngine
 {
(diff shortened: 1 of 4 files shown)


src/TSMapEditor/CCEngine/Theater.cs
```diff
@@ -75,6 +79,11 @@ public Theater(string name)
         public string FileExtension { get; set; }
         public char NewTheaterBuildingLetter { get; set; }
 
+        public List<string> RoughConnectToTileSets { get; set; }
+        public List<string> SandConnectToTileSets { get; set; }
+        public List<string> PaveConnectToTileSets { get; set; }
+        public List<string> GreenConnectToTileSets { get; set; }
```
Rampastring, line 86: Is there a reason why these exist as properties instead of just being handled like for the generic ground types, iow. read from the key at line 187?

src/TSMapEditor/Mutations/Mutation.cs
```diff
@@ -134,6 +134,9 @@ protected void ApplyGenericAutoLAT(int minX, int minY, int maxX, int maxY)
                     // alt. terrain
                     // For example, ~~~Snow shouldn't be auto-LAT'd when it's next to a tile belonging to ~~~Straight Dirt Roads
 
+                    var autoLatGround = latGrounds.Find(g => (g.GroundTileSet.Index == tileSetIndex || g.TransitionTileSet.Index == tileSetIndex) &&
+                    g.TransitionTileSet.Index != baseTileSetId && g.BaseTileSet.Index == baseTileSetId);
```
Rampastring: Incorrect indentation - this is part of the statement that begins on the previous line so it should be intended 4 spaces more to the right

Rampastring (changes requested):

src/TSMapEditor/CCEngine/Theater.cs
Starkku: I did mention this in the PR text, it is to allow setting these without touching the game's theater configuration INI.

Rampastring: Right.

I think optimally these could be read either from Theaters.ini or the game's theater config INI, if not specified in Theaters.ini. Could you edit the PR to make that possible?

Starkku: Done.

Rampastring (approved):

PR #150, opened 2024-05-03
Change MIX files to be loaded from a single list

ZivDero:
Changes mix files to be loaded from one list with the value controlling if they are required or optional. Meant to help merge the 2 TS branches.

Config for YR:
```
0=langmd.mix,1
1=language.mix,1
2=$EXPANDMD
3=ra2md.mix,1
4=ra2.mix,1
5=cachemd.mix,1
6=cache.mix,1
7=localmd.mix,1
8=local.mix,1
9=$RA2ECACHE
10=$RA2ELOCAL
11=conqmd.mix,1
12=genermd.mix,1
13=generic.mix,1
14=isogenmd.mix,1
15=isogen.mix,1
16=conquer.mix,1
17=marble.mix,0		; borrowed from Final Alert 2
```

### DIFF
--- a/src/TSMapEditor/CCEngine/CCFileManager.cs
+++ b/src/TSMapEditor/CCEngine/CCFileManager.cs
@@ -2,6 +2,7 @@ using Rampastring.Tools;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using TSMapEditor.UI;
 
 namespace TSMapEditor.CCEngine
 {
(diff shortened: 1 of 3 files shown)


Rampastring (commented):
This looks somehow heavy to parse for my eyes...

I'd prefer something like

```ini
0=landmd.mix,1
1=languagemd.mix,1
2=$EXPANDMD
; ...
```

Where the optional number after the name would denote whether it's required

Maybe even the current style would be good enough if the "required" status could be specified with a boolean or a number though

```ini
langmd.mix=1
language.mix=1
$EXPANDMD=0
```

src/TSMapEditor/CCEngine/CCFileManager.cs
```diff
@@ -34,8 +35,30 @@ public void ReadConfig()
...
+                bool isRequired = false;
+
+                if (parts.Length > 1)
+                    isRequired = Conversions.BooleanFromString(parts[1], isRequired);
+
+                if (isRequired)
+                    LoadRequiredMixFile(mixName);
+                else
+                    LoadOptionalMixFile(mixName);
+            }
```
Rampastring: Make this part into its own function, such as `LoadMixFiles`, to keep the general `ReadConfig` function flow clean. I liked how it was originally just a few `DoForEveryValueInSection` calls.

It could even be `LoadMixFile(string entry)` so it could be called with `iniFile.DoForEveryValueInSection`.

Rampastring (changes requested):

Rampastring (approved):

PR #151, opened 2024-05-03
Fix veins' palette

ZivDero:
Veins and the Veinhole monster are set to use unit palette in TS. In RA2/YR, this is still the case for the Veinhole monster, but veins themselves now use theater palette. This PR addresses this issue.

### DIFF
--- a/src/TSMapEditor/Rendering/TheaterGraphics.cs
+++ b/src/TSMapEditor/Rendering/TheaterGraphics.cs
@@ -1294,12 +1294,11 @@ namespace TSMapEditor.Rendering
                     var shpFile = new ShpFile(loadedShpName);
                     shpFile.ParseFromBuffer(shpData);
                     XNAPalette palette = theaterPalette;
-
-                    if (overlayType.Tiberium || overlayType.IsVeins)
-                        palette = Constants.TheaterPaletteForTiberium ? tiberiumPalette : unitPalette;
-
+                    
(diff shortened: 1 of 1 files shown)


src/TSMapEditor/Rendering/TheaterGraphics.cs
```diff
@@ -1296,15 +1295,10 @@ public void ReadOverlayTextures(List<OverlayType> overlayTypes)
...
-
-                        if (Constants.TheaterPaletteForTiberium)
-                            palette = tiberiumPalette ?? theaterPalette;
-                    }
+                    if (overlayType.Tiberium || overlayType.IsVeins)
+                        palette = Constants.TheaterPaletteForTiberium ? tiberiumPalette : unitPalette;
 
-                    if (overlayType.Wall || overlayType.IsVeins)
+                    if (overlayType.Wall || overlayType.IsVeinholeMonster)
                         palette = unitPalette;
```
Rampastring, line 1299: Currently this seems confusing for veinhole monsters at first glance, considering they have both `IsVeins=yes` and `IsVeinholeMonster=yes`.

I think this would be clearer like

```cs
                    if (overlayType.Wall || overlayType.IsVeinholeMonster)
                        palette = unitPalette;
                    else if (overlayType.Tiberium || overlayType.IsVeins)
                        palette = Constants.TheaterPaletteForTiberium ? tiberiumPalette : unitPalette;
```

Rampastring (approved):

PR #155, opened 2024-05-07
Add a sound selection window for triggers

ZivDero:
- Read sounds from sound.ini (name in Constants.ini)
- Add a window to select sounds for triggers
- Refactor Themes and EvaSpeeches a bit to use ImmutableList and to be more consistent with each other
- Refactor Speeches for TS to be the same as the YR ones

### DIFF
--- a/src/TSMapEditor/Misc/ListExtensions.cs
+++ b/src/TSMapEditor/Misc/ListExtensions.cs
@@ -1,5 +1,4 @@
 ﻿using System.Collections.Generic;
-using System.Collections.Immutable;
 
 namespace TSMapEditor.Misc
 {
@@ -17,19 +16,7 @@ namespace TSMapEditor.Misc
         /// Fetches an element at the given index.
         /// If the element is out of bounds, returns null.
         /// </summary>
-        public static T GetElementIfInRange<T>(this List<T> list, int index)
-        {
-            if (index < 0 || index >= list.Count)
-                return default;
-
-            return list[index];
-        }
-
-        /// <summary>
-        /// Fetches an element at the given index.
-        /// If the element is out of bounds, returns null.
-        /// </summary>
-        public static T GetElementIfInRange<T>(this ImmutableList<T> list, int index)
+        public static T GetElementIfInRange<T>(this IList<T> list, int index)
         {
             if (index < 0 || index >= list.Count)
                 return default;
--- a/src/TSMapEditor/Misc/ListExtensions.cs
+++ b/src/TSMapEditor/Misc/ListExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Collections.Immutable;
 
 namespace TSMapEditor.Misc
 {
@@ -17,6 +18,18 @@ namespace TSMapEditor.Misc
         /// If the element is out of bounds, returns null.
         /// </summary>
         public static T GetElementIfInRange<T>(this List<T> list, int index)
+        {
+            if (index < 0 || index >= list.Count)
+                return default;
+
+            return list[index];
+        }
+
+        /// <summary>
+        /// Fetches an element at the given index.
+        /// If the element is out of bounds, returns null.
+        /// </summary>
+        public static T GetElementIfInRange<T>(this ImmutableList<T> list, int index)
         {
             if (index < 0 || index >= list.Count)
                 return default;
--- a/src/TSMapEditor/Models/Themes.cs
+++ b/src/TSMapEditor/Models/Themes.cs
@@ -20,12 +20,12 @@ namespace TSMapEditor.Models
         [INI(false)]
         public int Index { get; }
 
-        public string Name { get; } = string.Empty;
-        public double Length { get; }
-        public bool Normal { get; }
-        public int Scenario { get; }
-        public int Side { get; }
-        public bool Repeat { get; }
+        public string Name { get; set; } = string.Empty;
+        public double Length { get; set; }
+        public bool Normal { get; set; }
+        public int Scenario { get; set; }
+        public int Side { get; set; }
+        public bool Repeat { get; set; }
 
         public override string ToString()
         {
--- a/src/TSMapEditor/Models/Themes.cs
+++ b/src/TSMapEditor/Models/Themes.cs
@@ -19,12 +19,12 @@ namespace TSMapEditor.Models
         [INI(false)]
         public int Index { get; }
 
-        public string Name { get; set; } = string.Empty;
-        public double Length { get; set; }
-        public bool Normal { get; set; }
-        public int Scenario { get; set; }
-        public int Side { get; set; }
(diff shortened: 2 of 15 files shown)


src/TSMapEditor/Models/Themes.cs
```diff
@@ -19,12 +19,12 @@ public Theme(string iniName, int index)
...
-        public bool Normal { get; set; }
-        public int Scenario { get; set; }
-        public int Side { get; set; }
-        public bool Repeat { get; set; }
+        public string Name { get; private set; } = string.Empty;
+        public double Length { get; private set; }
+        public bool Normal { get; private set; }
+        public int Scenario { get; private set; }
+        public int Side { get; private set; }
+        public bool Repeat { get; private set; }
```
Rampastring: `INIDefineable` only looks through public setters. In other words, this breaks automatic setting of these properties based on Theme.ini data.

src/TSMapEditor/Models/Themes.cs
```diff
@@ -38,15 +44,21 @@ public Themes(IniFileEx themeIni)
...
+        public Theme Get(int index)
+        {
+            return List.Find(theme => theme.Index == index);
+        }
 
-        public Theme GetByIndex(int index) => (index < 0 || index >= themes.Count) ? null : themes[index];
+        public Theme Get(string name)
+        {
+            return List.Find(theme => theme.Name == name);
+        }
```
Rampastring, line 58: Think of all the wasted CPU cycles and the resulting increase in global warming that this change will cause :(

More seriously though, what is the point of this change? If you think the latter looks nicer (which it does), it could be done while still retaining the array-based access by changing it to `return List.GetElementIfInRange(index)`.

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1041,16 +1056,31 @@ private void StringDarkeningPanel_Hidden(object sender, EventArgs e)
...
             if (Constants.IsRA2YR)
             {
-                AssignParamValue(selectSpeechWindow.IsForEvent, map.Rules.EvaSpeeches.Speeches[selectSpeechWindow.SelectedObject].Name);
+                AssignParamValue(selectSpeechWindow.IsForEvent, selectSpeechWindow.SelectedObject.Name);
             }
             else
             {
-                AssignParamValue(selectSpeechWindow.IsForEvent, selectSpeechWindow.SelectedObject);
+                AssignParamValue(selectSpeechWindow.IsForEvent, selectSpeechWindow.SelectedObject.Index);
+            }
```
Rampastring: Can't make this an automatic suggestion due to GH limitations, but

```cs
var speech = selectSpeechWindow.SelectedObject;
AssignParamValue(selectSpeechWindow.IsForEvent, Constants.IsRA2YR ? speech.Name : speech.Index;
```

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1041,16 +1056,31 @@ private void StringDarkeningPanel_Hidden(object sender, EventArgs e)
...
+                return;
+
+            if (Constants.IsRA2YR)
+            {
+                AssignParamValue(selectSoundWindow.IsForEvent, selectSoundWindow.SelectedObject.Name);
+            }
+            else
+            {
+                AssignParamValue(selectSoundWindow.IsForEvent, selectSoundWindow.SelectedObject.Index);
             }
```
Rampastring:
```suggestion
            var sound = selectSoundWindow.SelectedObject;
            AssignParamValue(selectSoundWindow.IsForEvent, Constants.IsRA2YR ? sound.Name : sound.Index);
```

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1955,24 +1985,52 @@ private string GetParamValueText(string paramValue, TriggerParamType paramType,
...
                             return paramValue;
 
-                        if (!map.EditorConfig.Speeches.ContainsKey(intValue))
+                        speech = map.EditorConfig.Speeches.Get(intValue);
+
+                        if (speech == null)
                             return intValue + " - unknown speech";
 
-                        return intValue + " " + map.EditorConfig.Speeches[intValue];
+                        return speech.Name;
```
Rampastring: Since the index is relevant in TS, it should also be displayed.

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1955,24 +1985,52 @@ private string GetParamValueText(string paramValue, TriggerParamType paramType,
...
+                    {
+                        if (!intParseSuccess)
+                            return paramValue;
+
+                        sound = map.Rules.Sounds.Get(intValue);
+
+                        if (sound == null)
+                            return intValue + " - unknown sound";
+
+                        return sound.Name;
```
Rampastring: Since the index is relevant in TS, it should also be displayed.

Rampastring (changes requested):

src/TSMapEditor/Models/Themes.cs
ZivDero: It was indeed for niceness; I guess it is also technically different, but the index should in theory always match with the actual index, so w/e.

src/TSMapEditor/UI/Windows/TriggersWindow.cs
ZivDero: I wish it worked this way. Cannot resolve method.

Rampastring: `speech.Index.ToString(CultureInfo.InvariantCulture)` should be OK I think?

ZivDero: Yeah that works

src/TSMapEditor/Misc/ListExtensions.cs
```diff
@@ -23,6 +24,18 @@ public static T GetElementIfInRange<T>(this List<T> list, int index)
...
+        /// Fetches an element at the given index.
+        /// If the element is out of bounds, returns null.
+        /// </summary>
+        public static T GetElementIfInRange<T>(this ImmutableList<T> list, int index)
+        {
+            if (index < 0 || index >= list.Count)
+                return default;
+
+            return list[index];
+        }
```
Rampastring: Duplicate code. Could remove this one and change the method above to take an `IList<T>` instead of a concrete list implementation.

Rampastring (changes requested):

src/TSMapEditor/Misc/ListExtensions.cs
ZivDero: Yeah had such a thought a bit later

Rampastring (approved):

PR #158, opened 2024-05-10
Cliff Drawer

ZivDero:
Cliff Drawing tool. Currently includes YR config and thus targets YR branch, will be retargeted before merge.

### DIFF
--- a/src/TSMapEditor/Models/CliffType.cs
+++ b/src/TSMapEditor/Models/CliffType.cs
@@ -2,6 +2,7 @@
 using Rampastring.Tools;
 using System;
 using System.Collections.Generic;
+using System.Drawing.Drawing2D;
 using System.Linq;
 using System.Text.RegularExpressions;
 using TSMapEditor.GameMath;
@@ -46,7 +47,11 @@ namespace TSMapEditor.Models
                     var exit = tile.GetExit(neighbor.cp.Index);
                     exit = (CliffConnectionPoint)exit.MemberwiseClone();
 
-                    neighbors.Add(new CliffAStarNode(node, neighbor.cp, exit, placementCoords, tile));
+                    var newNode = new CliffAStarNode(node, neighbor.cp, exit, placementCoords, tile);
+
+                    // Make sure that the new node doesn't overlap anything
+                    if (newNode.OccupiedTiles.Count - node.OccupiedTiles.Count == newNode.Tile.Foundation.Count)
+                        neighbors.Add(newNode);
                 }
             }
             return neighbors;
@@ -85,6 +90,9 @@ namespace TSMapEditor.Models
             Entry = entry;
             Exit = exit;
             Destination = Parent.Destination;
+
+            OccupiedTiles = new HashSet<Vector2>(parent.OccupiedTiles);
+            OccupiedTiles.UnionWith(tile.Foundation.Select(coord => coord + Location));
         }
 
         public static CliffAStarNode MakeStartNode(Vector2 location, Vector2 destination, CliffSide startingSide)
@@ -161,7 +169,7 @@ namespace TSMapEditor.Models
                     result += Parent.GScore + Helpers.VectorDistance(Parent.ExitCoords, ExitCoords);
                     if (Entry != null && Parent.Exit.Side != Entry.Side)
                     {
-                        result *= 1.2f;
+                        //result *= 1.2f;
                     }
                 }
 
@@ -174,7 +182,7 @@ namespace TSMapEditor.Models
         public float FScore => GScore + HScore;
         public CliffAStarNode Parent;
 
-        public List<Vector2> OccupiedTiles = new List<Vector2>();
+        public HashSet<Vector2> OccupiedTiles = new HashSet<Vector2>();
     }
 
     public class CliffTile
@@ -182,6 +190,7 @@ namespace TSMapEditor.Models
         public string TileSet { get; set; }
         public int TileIndexInSet { get; set; }
         public List<CliffConnectionPoint> ConnectionPoints { get; set; }
+        public List<Vector2> Foundation { get; set; }
 
         public CliffConnectionPoint GetExit(int entryIndex)
         {
@@ -239,10 +248,21 @@ namespace TSMapEditor.Models
                     });
                 }
 
+                string foundationString = iniSection.GetStringValue("Foundation", string.Empty);
+                if (!Regex.IsMatch(foundationString, "(\\d+?,\\d+?;)*(\\d+?,\\d+?)"))
+                    throw new INIConfigException($"Cliff {sectionName} has an invalid Foundation {foundationString}!");
+
+                var foundation = foundationString.Split(";").Select(coordinateString =>
+                {
+                    var coordinateParts = coordinateString.Split(",");
+                    return new Vector2(int.Parse(coordinateParts[0]), int.Parse(coordinateParts[1]));
+                }).ToList();
+
                 Tiles.Add(new CliffTile()
                 {
                     ConnectionPoints = connectionPoints,
-                    TileIndexInSet = tileIndexInSet
+                    TileIndexInSet = tileIndexInSet,
+                    Foundation = foundation
                 });
             }
         }
--- a/src/TSMapEditor/Models/CliffType.cs
+++ b/src/TSMapEditor/Models/CliffType.cs
@@ -216,12 +216,12 @@ namespace TSMapEditor.Models
 
             string indicesString = iniSection.GetStringValue("TileIndices", null);
             if (indicesString == null || !Regex.IsMatch(indicesString, "^((?:\\d+?,)*(?:\\d+?))$"))
-                throw new INIConfigException($"Cliff {iniSection.SectionName} has invalid TileIndices list: {indicesString}!");
+                throw new INIConfigException($"Connected Tile {iniSection.SectionName} has invalid TileIndices list: {indicesString}!");
 
 
             string tileSet = iniSection.GetStringValue("TileSet", null);
             if (string.IsNullOrWhiteSpace(tileSet))
-                throw new INIConfigException($"Cliff {iniSection.SectionName} has no TileSet!");
+                throw new INIConfigException($"Connected Tile {iniSection.SectionName} has no TileSet!");
 
             TileSetName = tileSet;
 
@@ -233,13 +233,13 @@ namespace TSMapEditor.Models
             {
                 string coordsString = iniSection.GetStringValue($"ConnectionPoint{i}", null);
                 if (coordsString == null || !Regex.IsMatch(coordsString, "^\\d+?,\\d+?$"))
-                    throw new INIConfigException($"Cliff {iniSection.SectionName} has invalid ConnectionPoint{i} value: {coordsString}!");
+                    throw new INIConfigException($"Connected Tile {iniSection.SectionName} has invalid ConnectionPoint{i} value: {coordsString}!");
 
                 Point2D coords = Point2D.FromString(coordsString);
 
                 string directionsString = iniSection.GetStringValue($"ConnectionPoint{i}.Directions", null);
                 if (directionsString == null || directionsString.Length != (int)Direction.Count || Regex.IsMatch(directionsString, "[^01]"))
-                    throw new INIConfigException($"Cliff {iniSection.SectionName} has invalid ConnectionPoint{i}.Directions value: {directionsString}!");
+                    throw new INIConfigException($"Connected Tile {iniSection.SectionName} has invalid ConnectionPoint{i}.Directions value: {directionsString}!");
 
                 byte directions = Convert.ToByte(directionsString, 2);
 
@@ -248,7 +248,7 @@ namespace TSMapEditor.Models
                 {
                     "front" => CliffSide.Front,
                     "back" => CliffSide.Back,
-                    _ => throw new INIConfigException($"Cliff {iniSection.SectionName} has an invalid ConnectionPoint{i}.Side value: {sideString}!")
+                    _ => throw new INIConfigException($"Connected Tile {iniSection.SectionName} has an invalid ConnectionPoint{i}.Side value: {sideString}!")
                 };
 
                 int[] requiredTiles, forbiddenTiles;
@@ -283,7 +283,7 @@ namespace TSMapEditor.Models
 
             string foundationString = iniSection.GetStringValue("Foundation", string.Empty);
             if (!Regex.IsMatch(foundationString, "^((?:\\d+?,\\d+?\\|)*(?:\\d+?,\\d+?))$"))
-                throw new INIConfigException($"Cliff {iniSection.SectionName} has an invalid Foundation: {foundationString}!");
+                throw new INIConfigException($"Connected Tile {iniSection.SectionName} has an invalid Foundation: {foundationString}!");
 
             Foundation = foundationString.Split("|").Select(coordinateString => Point2D.FromString(coordinateString)).ToHashSet();
         }
@@ -350,6 +350,12 @@ namespace TSMapEditor.Models
                 var parts = sectionName.Split('.');
                 if (parts.Length != 2 || parts[0] != IniName || !int.TryParse(parts[1], out int index))
                     continue;
+
+                if (Tiles.Exists(tile => tile.Index == index))
+                {
+                    throw new INIConfigException(
+                        $"Connected Tile {iniName} has multiple tiles with the same index {index}!");
+                }
 
                 Tiles.Add(new CliffTile(iniFile.GetSection(sectionName), index));
             }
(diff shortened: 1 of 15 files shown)


src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        /// <summary>
+        /// Previous node
+        /// </summary>
+        public CliffAStarNode Parent;
+
+        /// <summary>
+        /// Accumulated set of all cell coordinates occupied up to this node
+        /// </summary>
+        public HashSet<Vector2> OccupiedCells = new HashSet<Vector2>();
```
Rampastring: Move these to the beginning of the class, below the constructors.

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+
+                    var exit = tile.GetExit(connectionPoint.Index);
+                    var newNode = new CliffAStarNode(this, exit, placementCoords, tile);
+
+                    // Make sure that the new node doesn't overlap anything
+                    if (newNode.OccupiedCells.Count - OccupiedCells.Count == newNode.Tile.Foundation.Count)
+                        neighbors.Add(newNode);
+                }
+            }
+            return neighbors;
```
Rampastring:
```suggestion
            var neighbors = new List<CliffAStarNode>();
            
            foreach (var (connectionPoint, directions) in possibleNeighbors)
            {
                if (connectionPoint.Side != Exit.Side)
                    continue;

                foreach (Direction dir in directions)
                {
                    Vector2 placementOffset = Helpers.VisualDirectionToPoint(dir).ToXNAVector() - connectionPoint.CoordinateOffset;
                    Vector2 placementCoords = ExitCoords + placementOffset;

                    var exit = tile.GetExit(connectionPoint.Index);
                    var newNode = new CliffAStarNode(this, exit, placementCoords, tile);

                    // Make sure that the new node doesn't overlap anything
                    if (newNode.OccupiedCells.Count - OccupiedCells.Count == newNode.Tile.Foundation.Count)
                        neighbors.Add(newNode);
                }
            }
            
            return neighbors;
```

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        /// <summary>
+        /// Tile's in-editor index
+        /// </summary>
+        public int Index { get; set; }
+
+        /// <summary>
+        /// Tile's Tile Set
+        /// </summary>
+        public string TileSet { get; set; }
```
Rampastring: This is named like it was a `TileSet` instance, but actually it's a string, the name of a TileSet. Rename it to `TileSetName` (at least in code; that's not necessary for the config).

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+            if (string.IsNullOrWhiteSpace(tileSet))
+                throw new INIConfigException($"Cliff {iniSection.SectionName} has no TileSet!");
+
+            TileSet = tileSet;
+
+            IndicesInTileSet = indicesString.Split(',').Select(int.Parse).ToList();
+
+            ConnectionPoints = new CliffConnectionPoint[2];
+
+            for (int i = 0; i < 2; i++)
```
Rampastring:
```suggestion
            for (int i = 0; i < ConnectionPoints.Length; i++)
```

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        /// <summary>
+        /// Places this tile connects to other tiles
+        /// </summary>
+        public CliffConnectionPoint[] ConnectionPoints { get; set; }
+
+        /// <summary>
+        /// Set of all relative cell coordinates this tile occupies
+        /// </summary>
+        public HashSet<Vector2> Foundation { get; set; }
```
Rampastring: Cell coordinates? Any reason why this uses the float-based `Vector2` instead of the integer-based `Point2D`?

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+                {
+                    "front" => CliffSide.Front,
+                    "back" => CliffSide.Back,
+                    _ => throw new INIConfigException($"Cliff {iniSection.SectionName} has an invalid ConnectionPoint{i}.Side value: {sideString}!")
+                };
+
+                int[] requiredTiles, forbiddenTiles;
+
+                var requiredTilesList =
+                    iniSection.GetListValue($"ConnectionPoint{i}.RequiredTiles", ',', int.Parse);
```
Rampastring, line 257: You might run into culture issues by just calling `int.Parse`. Doing something like `s => int.Parse(s, CultureInfo.InvariantCulture)` would avoid it.

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+
+                if (requiredTilesList.Count > 0)
+                {
+                    requiredTiles = requiredTilesList.ToArray();
+                    forbiddenTiles = Array.Empty<int>();
+                }
+                else
+                {
+                    var forbiddenTilesList =
+                        iniSection.GetListValue($"ConnectionPoint{i}.ForbiddenTiles", ',', int.Parse);
```
Rampastring, line 267: You might run into culture issues by just calling `int.Parse`. Doing something like `s => int.Parse(s, CultureInfo.InvariantCulture)` would avoid it.

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+    {
+        /// <summary>
+        /// Index of the connection point, 0 or 1
+        /// </summary>
+        public int Index { get; set; }
+
+        /// <summary>
+        /// Offset of this connection point relative to the tile's (0,0) point
+        /// </summary>
+        public Vector2 CoordinateOffset { get; set; }
```
Rampastring: Any reason why this uses the float-based `Vector2` instead of the integer-based `Point2D`?

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,362 @@
...
+
+namespace TSMapEditor.Models
+{
+    public enum CliffSide
+    {
+        Front,
+        Back
+    }
+
+    public struct CliffConnectionPoint
```
Rampastring: This is a non-immutable struct, iow. a struct with setters. It is a common way to enable one to shoot themselves in the foot and should be avoided unless there's a good (typically performance) reason.

If there is never a need to assign a value to the properties after an instance of this struct has been created, the members should be made read-only instead.

It could also be that there is no need for this to be a struct rather than class. Having a struct instead of class can help with performance if the struct is small enough and especially if there's a need to create instances often, but at 33 bytes in size, this struct is on the edge of being too large for performance gains.

src/TSMapEditor/Models/EditorConfig.cs
```diff
@@ -394,5 +396,24 @@ private void ReadSpeeches()
 
             Speeches = new EvaSpeeches(speeches.ToArray());
         }
+
+        private void ReadCliffs()
+        {
+            ConnectedOverlays.Clear();
+
+            var iniFile = new IniFile(Environment.CurrentDirectory + "/Config/CliffDrawer.ini");
```
Rampastring: I wonder if we should call this file something else to begin with since we know that this can be used for stuff other than cliffs, and similarly also rename the section (`[Cliffs]`).

ConnectedTileDrawer?

To clarify, I'm not asking to rename all the internal references to cliffs, but renaming the user/modder-facing elements is best to do before people start creating configs for shores, roads and the like.

src/TSMapEditor/Mutations/Classes/DrawCliffMutation.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// A mutation for drawing cliffs.
+    /// </summary>
+    public class DrawCliffMutation : Mutation
+    {
+        public DrawCliffMutation(IMutationTarget mutationTarget, List<Point2D> cliffPath, CliffType cliffType, CliffSide startingSide, int randomSeed, byte extraHeight) : base(mutationTarget)
+        {
+            if (cliffPath.Count < 2)
+            {
+                throw new ArgumentException(nameof(DrawCliffMutation) +
+                                            ": to draw a cliff at least 2 path vertices are required.");
```
Rampastring:
```suggestion
                throw new ArgumentException(nameof(DrawCliffMutation) +
                    ": to draw a cliff at least 2 path vertices are required.");
```

src/TSMapEditor/UI/CursorActions/DrawCliffCursorAction.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            Rectangle textBackgroundRectangle = new Rectangle((int)textPosition.X - Constants.UIEmptySideSpace,
+                (int)textPosition.Y - Constants.UIEmptyTopSpace,
+                (int)textDimensions.X + Constants.UIEmptySideSpace * 2,
+                (int)textDimensions.Y + Constants.UIEmptyBottomSpace + Constants.UIEmptyTopSpace);
+
+            Renderer.FillRectangle(textBackgroundRectangle, UISettings.ActiveSettings.PanelBackgroundColor);
+            Renderer.DrawRectangle(textBackgroundRectangle, UISettings.ActiveSettings.PanelBorderColor);
+
+            Renderer.DrawStringWithShadow(text, Constants.UIBoldFont, textPosition, Color.Yellow);
```
Rampastring: Instead of duplicating this code, you could call `base.DrawText`.

src/TSMapEditor/UI/CursorActions/DrawCliffCursorAction.cs
```diff
@@ -0,0 +1,187 @@
...
+        public override bool DrawCellCursor => true;
+
+        private readonly CliffType cliffType;
+
+        private List<Point2D> cliffPath;
+        private CliffSide cliffSide = CliffSide.Front;
+        private DrawCliffMutation previewMutation;
+        private byte extraHeight = 0;
+
+        private readonly int randomSeed = Guid.NewGuid().GetHashCode();
```
Rampastring: Not that it matters a lot, but this seems a bit overkill.

```suggestion
        private readonly int randomSeed = new Random().Next();
```

src/TSMapEditor/UI/TopBar/TopBarMenu.cs
```diff
@@ -127,6 +133,22 @@ public override void Initialize()
...
+            if (cliffCount > 0)
+            {
+                if (cliffCount == 1)
+                {
+                    editContextMenu.AddItem("Draw Cliff", () => mapView.EditorState.CursorAction =
+                        new DrawCliffCursorAction(mapView, theaterMatchingCliffs[0]), null, null, null);
+                }
+                else
+                {
+                    editContextMenu.AddItem("Draw Cliff...", SelectCliff, null, null, null);
```
Rampastring:
```suggestion
                    editContextMenu.AddItem("Draw Cliff...", () => selectCliffWindow.Open(), null, null, null);
```

Then the `SelectCliff` method can be removed.

src/TSMapEditor/UI/Windows/SelectCliffWindow.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    public class SelectCliffWindow : SelectObjectWindow<CliffType>
+    {
+        public SelectCliffWindow(WindowManager windowManager, Map map) : base(windowManager)
+        {
+            this.map = map;
+        }
+
+        private readonly Map map;
+        public bool Success = false;
```
Rampastring: What is this needed for? Isn't checking for SelectedObject to be non-null enough?

I believe the current implementation also breaks selecting a cliff type by double-clicking on one in the list, since double-clicking on a member of the list does not set this property to `true`.

Rampastring (changes requested):
Generally looks good. Some things to clarify and/or fix.

src/TSMapEditor/Models/CliffType.cs
ZivDero: Convenience; the main reason is because I want normal Euclidean distance as a float later on, and Point2D seems to be more focused on cell math and in integers only.

Rampastring: It's better to have fitting types everywhere and instead perform a conversion where you calculate that Euclidean distance.

src/TSMapEditor/Models/EditorConfig.cs
ZivDero: I was also thinking it may be weird to have shores or smth under "Draw Cliff..."

Rampastring: Could also rename that to "Draw Connected Tiles..."

src/TSMapEditor/UI/Windows/SelectCliffWindow.cs
ZivDero: The issue is that closing with X is the same as closing with "select", so if you click on something and click X, it registers as having selected something, even though the user clearly intended to cancel the action.

Rampastring: The `SelectObjectWindow` base class has this functionality that should prevent that from being the case

```cs
            // Make pressing X not save changes
            if (btnClose != null)
                btnClose.LeftClick += (s, e) => SelectedObject = initialSelection;
```

Even if it was non-functional, it should be fixed at the base class level, not with window-specific functionality.

ZivDero: I was not aware it had such functionality. I'll remove this then, but I am not 100% certain it works (but it may)

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,368 @@
...
+            ConnectionPoints = new CliffConnectionPoint[2];
+
+            for (int i = 0; i < ConnectionPoints.Length; i++)
+            {
+                string coordsString = iniSection.GetStringValue($"ConnectionPoint{i}", null);
+                if (coordsString == null || !Regex.IsMatch(coordsString, "^\\d+?,\\d+?$"))
+                    throw new INIConfigException($"Cliff {iniSection.SectionName} has invalid ConnectionPoint{i} value: {coordsString}!");
+
+                var coordParts = coordsString.Split(',').Select(s => int.Parse(s, CultureInfo.InvariantCulture)).ToList();
+                Point2D coords = new Point2D(coordParts[0], coordParts[1]);
```
Rampastring:
```suggestion
                Point2D coords = Point2D.FromString(coordsString);
```

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,368 @@
...
+
+            string foundationString = iniSection.GetStringValue("Foundation", string.Empty);
+            if (!Regex.IsMatch(foundationString, "^((?:\\d+?,\\d+?\\|)*(?:\\d+?,\\d+?))$"))
+                throw new INIConfigException($"Cliff {iniSection.SectionName} has an invalid Foundation: {foundationString}!");
+
+            Foundation = foundationString.Split("|").Select(coordinateString =>
+            {
+                var coordinateParts = coordinateString.Split(",");
+                return new Point2D(int.Parse(coordinateParts[0]), int.Parse(coordinateParts[1]));
+            }).ToHashSet();
```
Rampastring:
```suggestion
            Foundation = foundationString.Split("|").Select(coordinateString => Point2D.FromString(coordinateString)).ToHashSet();
```

src/TSMapEditor/Models/CliffType.cs
```diff
@@ -0,0 +1,368 @@
...
+
+            Tiles = new List<CliffTile>();
+
+            foreach (var sectionName in iniFile.GetSections())
+            {
+                var parts = sectionName.Split('.');
+                if (parts.Length != 2 || parts[0] != IniName || !int.TryParse(parts[1], out int index))
+                    continue;
+
+                Tiles.Add(new CliffTile(iniFile.GetSection(sectionName), index));
```
Rampastring, line 360: Should there be a check to disallow two items from having a duplicate index?

src/TSMapEditor/Models/EditorConfig.cs
```diff
@@ -394,5 +396,24 @@ private void ReadSpeeches()
 
             Speeches = new EvaSpeeches(speeches.ToArray());
         }
+
+        private void ReadCliffs()
+        {
+            ConnectedOverlays.Clear();
```
Rampastring: Copy/paste mistake

src/TSMapEditor/Mutations/Classes/DrawCliffMutation.cs
```diff
@@ -0,0 +1,185 @@
...
+        {
+            if (tile == null)
+                return;
+
+            for (int i = 0; i < tile.TMPImages.Length; i++)
+            {
+                MGTMPImage image = tile.TMPImages[i];
+                if (image.TmpImage == null)
+                    continue;
+                int cx = targetCellCoords.X + i % tile.Width;
```
Rampastring:
```suggestion
                    continue;

                int cx = targetCellCoords.X + i % tile.Width;
```

Rampastring (changes requested):
A few more.

Rampastring (approved):

PR #163, opened 2024-05-28
Add shores and roads to YR config

ZivDero:


Rampastring:
This has been manually merged into `yr`.

PR #167, opened 2024-05-28
Add option to add a modifier to distance calculation in connected tile drawer

ZivDero:


src/TSMapEditor/Models/CliffType.cs
```diff
@@ -108,7 +108,7 @@ public CliffAStarNode(CliffAStarNode parent, CliffConnectionPoint exit, Point2D
         /// Distance to end node
         /// </summary>
         public float HScore => Vector2.Distance(Destination.ToXNAVector(), ExitCoords.ToXNAVector());
-        public float FScore => GScore * 0.7f + HScore;
+        public float FScore => GScore * 0.7f + HScore + Tile?.DistanceModifier ?? 0;
```
Rampastring: I think it would be good to clarify the operation order here.

```suggestion
        public float FScore => GScore * 0.7f + HScore + (Tile?.DistanceModifier ?? 0);
```

src/TSMapEditor/Config/ConnectedTileDrawer.ini
```diff
@@ -19,6 +19,8 @@
 ; ConnectionPoint0/1.ForbiddenTiles=IDs of tiles this point cannot connect to, optional, RequiredTiles take priority
 ; Foundation=0,0|0,1|1,1|1,2 - optional, list of coordinates this tile occupies relative to its 0,0 point. If not specified, automatically filled by editor based on the shape of the tile
 ; ExtraPriority=integer - secondary sorting key when all else is equal, higher values mean higher priority
+; DistanceModifier=integer - value to add to the primary key when sorting tiles, negative means the tile is "better" or "closer to the target at a lower cost".
+                           ; Use with caution as it may produce undesireable results, or break the pathfinding in the worst case!
```
Rampastring: There's nothing behind the comment, no reason to leave uncommented space around :P

```suggestion
;                            Use with caution as it may produce undesireable results, or break the pathfinding in the worst case!
```

Rampastring (changes requested):

Rampastring (approved):

PR #173, opened 2024-06-07
Improve lighting calculations

ZivDero:
- Buildings now take their foundation into account properly when lighting
- Adding/removing/moving buildings now only updates the cells they light, not the whole map, reducing lag

### DIFF
--- a/src/TSMapEditor/Models/Structure.cs
+++ b/src/TSMapEditor/Models/Structure.cs
@@ -197,7 +197,7 @@ namespace TSMapEditor.Models
                             continue;
 
                         if (!litTiles.ContainsKey(tile) ||
-                            litTiles.ContainsKey(tile) && litTiles[tile] > distanceInLeptons)
+                            (litTiles.ContainsKey(tile) && litTiles[tile] > distanceInLeptons))
                         {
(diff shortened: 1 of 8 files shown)


src/TSMapEditor/Models/Structure.cs
```diff
@@ -155,6 +155,74 @@ public override byte Facing
...
+                        int yDifference = centerPosition.Y - y;
+
+                        double distanceInCells = Math.Sqrt(xDifference * xDifference + yDifference * yDifference);
+                        double distanceInLeptons = distanceInCells * Constants.CellSizeInLeptons;
+
+                        if (distanceInLeptons > ObjectType.LightVisibility)
+                            continue;
+
+                        if (!litTiles.ContainsKey(tile) ||
+                            litTiles.ContainsKey(tile) && litTiles[tile] > distanceInLeptons)
```
Rampastring:
```suggestion
                            (litTiles.ContainsKey(tile) && litTiles[tile] > distanceInLeptons))
```

Clarifying the intended operator order.

Rampastring (changes requested):

Rampastring (approved):

PR #175, opened 2024-06-11
Allow parsing script action options from another section

Starkku:
Allows using `OptionsSectionName` in `ScriptActions.ini` to allow reading the preset options (`OptionN` key) from another section. Practically necessary for certain Phobos AI script actions because several of them use the same fixed list of possible parameter values, which would be exceedingly verbose to define separately for each of them.

Because the editor currently makes assumption that every section in this INI file is a script action, I made it so that ones prefixed with `$` are not treated as such so that they can be used purely for defining the options (you can also just point to another script action I guess). A 'proper' solution to this would have required multiple parsing passes of the INI file to figure out which sections are actually used in `OptionsSectionName` or an equally complicated solution, or having it expect the sections in a specific order. If anyone can think of a better way to do this, I can adjust it.

### DIFF
--- a/src/TSMapEditor/CCEngine/ScriptAction.cs
+++ b/src/TSMapEditor/CCEngine/ScriptAction.cs
@@ -33,18 +33,29 @@ namespace TSMapEditor.CCEngine
         public string Name { get; set; } = "Unknown action";
         public string Description { get; set; } = "No description";
         public string ParamDescription { get; set; } = "Use 0";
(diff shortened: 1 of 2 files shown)


Rampastring (approved):
An option for avoiding the `$` syntax would be reading all script actions from a definitions section, like the games have for TechnoTypes (`BuildingTypes` etc.).

This shall suffice at least for now though.

PR #176, opened 2024-06-11
Allow script actions to use window for parameter selection

Starkku:
Pretty much what it says in the title, allows setting `UseWindowSelection` for script actions in `ScriptActions.ini` to use a selector window for the preset options (`OptionN`) key. Particularly useful (and needed) for certain Phobos script actions that use a list of values that far exceeds what the context menu can show.

A different button is displayed for opening the window on script actions that use it, from the one that opens the context menu.

### DIFF
--- a/src/TSMapEditor/CCEngine/ScriptAction.cs
+++ b/src/TSMapEditor/CCEngine/ScriptAction.cs
@@ -1,11 +1,12 @@
-﻿using Rampastring.Tools;
+﻿using Microsoft.Xna.Framework;
+using Rampastring.Tools;
 using System;
 using System.Collections.Generic;
 using TSMapEditor.Models.Enums;
 
 namespace TSMapEditor.CCEngine
 {
-    public struct ScriptActionPresetOption
+    public class ScriptActionPresetOption
     {
         public int Value;
         public string Text;
@@ -18,7 +19,10 @@ namespace TSMapEditor.CCEngine
 
         public string GetOptionText()
         {
-            return Value + " - " + Text;
+            if (string.IsNullOrEmpty(Text))
+                return Value.ToString();
+            else
+                return Value + " - " + Text;
         }
     }
 
@@ -36,6 +40,7 @@ namespace TSMapEditor.CCEngine
         public string OptionsSectionName { get; set; } = string.Empty;
         public TriggerParamType ParamType { get; set; } = TriggerParamType.Unknown;
         public List<ScriptActionPresetOption> PresetOptions { get; } = new List<ScriptActionPresetOption>(0);
+        public bool UseWindowSelection { get; set; } = false;
 
         public void ReadIniSection(IniFile iniFile, string sectionName)
         {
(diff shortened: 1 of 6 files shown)


src/TSMapEditor/Config/UI/Windows/ScriptsWindow.ini
```diff
@@ -25,6 +25,7 @@ $CC20=tbParameterValue:EditorNumberTextBox
 $CC21=btnEditorPresetValues:MenuButton
 $CC22=lblActionDescription:XNALabel
 $CC23=panelActionDescription:EditorPanel
+$CC24=btnEditorPresetValuesWindow:EditorButton
```
Rampastring:
```suggestion
$CC21=btnEditorPresetValues:MenuButton
$CC22=btnEditorPresetValuesWindow:EditorButton
$CC23=lblActionDescription:XNALabel
$CC24=panelActionDescription:EditorPanel
```

Since this exists visually in the same space as `btnEditorPresetValues`, and the `[btnEditorPresetValuesWindow]` section has been placed under the `[btnEditorPresetValues]` section, it would be logical for this to also be defined right below `btnEditorPresetValues` in the list.

src/TSMapEditor/UI/Windows/ScriptsWindow.cs
```diff
@@ -216,6 +226,18 @@ private void BtnEditorPresetValues_LeftClick(object sender, EventArgs e)
             }
         }
 
+        private void BtnEditorPresetValuesWindow_LeftClick(object sender, EventArgs e)
+        {
+            if (editedScript == null)
+                return;
+
+            if (lbActions.SelectedItem == null)
+                return;
```
Rampastring, line 313:
```suggestion
            if (editedScript == null || lbActions.SelectedItem == null)
                return;
```

src/TSMapEditor/UI/Windows/SelectScriptActionPresetOptionWindow.cs
```diff
@@ -0,0 +1,83 @@
...
+        protected override void ListObjects()
+        {
+            lbObjectList.Clear();
+
+            foreach (ScriptActionPresetOption presetOption in presetOptions)
+            {
+                var item = new XNAListBoxItem() { Text = $"{presetOption.GetOptionText()}", Tag = presetOption };
+
+                if (presetOption.Color != null)
+                    item.TextColor = (Color)presetOption.Color;
```
Rampastring:
```suggestion
                if (presetOption.Color.HasValue)
                    item.TextColor = presetOption.Color.Value;
```

Not really a complaint, I just prefer this syntax when working with nullable value types.

src/TSMapEditor/CCEngine/ScriptAction.cs
```diff
@@ -1,24 +1,34 @@
...
 using TSMapEditor.Models.Enums;
 
 namespace TSMapEditor.CCEngine
 {
-    public struct ScriptActionPresetOption
+    public class ScriptActionPresetOption
     {
         public int Value;
         public string Text;
+        public Color? Color;
```
Rampastring: As far as I can see, there is currently no way to actually assign this a value? The INI parsing code for preset options is unchanged.

src/TSMapEditor/CCEngine/ScriptAction.cs
```diff
@@ -1,24 +1,34 @@
...
 
-        public ScriptActionPresetOption(int value, string text)
+        public ScriptActionPresetOption()
+        {
+        }
```
Rampastring: Unless I'm missing something, this constructor is never used, so it can be removed.

Rampastring (changes requested):

Starkku:
Rebased and feedback addressed.

Rampastring (approved):

PR #177, opened 2024-06-11
Add superweapon selector window and superweapon ID param type

Starkku:
Changes superweapon parameter for trigger events & actions to use the window-based selector, as most YR projects tend to have much larger selection of superweapons available which exceeds what the context menu can comfortably show. I was initially going to make this optional and have default behaviour use the context menu still, but I figured that might make the code unnecessarily complicated. If that is preferrable however, I can make it happen.

Additionally added support for superweapon ID (section name) as a parameter type, which is used by one trigger event added by Ares.

### DIFF
--- a/src/TSMapEditor/UI/Windows/TriggersWindow.cs
+++ b/src/TSMapEditor/UI/Windows/TriggersWindow.cs
@@ -102,6 +102,7 @@ namespace TSMapEditor.UI.Windows
         private SelectStringWindow selectStringWindow;
         private SelectSpeechWindow selectSpeechWindow;
         private SelectSoundWindow selectSoundWindow;
+        private SelectSuperWeaponTypeWindow selectSuperWeaponTypeWindow;
         private SelectParticleSystemTypeWindow selectParticleSystemTypeWindow;
         private CreateRandomTriggerSetWindow createRandomTriggerSetWindow;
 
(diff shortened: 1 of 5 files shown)


src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -2041,6 +2074,13 @@ private string GetParamValueText(string paramValue, TriggerParamType paramType,
                         return intValue + " - nonexistent super weapon";
 
                     return intValue + " " + map.Rules.SuperWeaponTypes[intValue].GetDisplayStringWithoutIndex();
+                case TriggerParamType.SuperWeaponID:
+                    var swType = map.Rules.SuperWeaponTypes.Find(sw => sw.ININame.Equals(paramValue, StringComparison.Ordinal));
+
+                    if (swType == null)
+                        return paramValue;
+
+                    return swType.GetDisplayStringWithoutIndex();
```
ZivDero, line 2294: How about just `.ToString()`?

Starkku: This predated my changes and given that most elegant way of changing it (overriding `ToString()` to return `ININame`) is not easibly doable as blanket case for all INI-defined types since `ININame` comes from an inteface and idt interfaces allow for method overrides for obvious reasons, I have elected to leave it alone for now. If someone else wants to refactor it afterwards be my guest.

Starkku:
Rebased and altered some minor things like the name of the param type to match changes introduced in between.

Rampastring (approved):

PR #178, opened 2024-06-11
Allow fill terrain modifier hotkey to apply to raise/lower cells

Starkku:
Allows holding the `Fill Terrain` hotkey while using the `Raise / Lower Individual Cells` tools to apply the action to an enclosed area similar to filling an area of cells with 1x1 tiles.

I was initially going to add its own customizable hotkey for this, but for sake of convenience and consistency I would have wanted both actions to use the same key, and current system does not allow mapping same key to multiple actions (for obvious reasons). So it was either this or hardcoding it to holding Ctrl or something.

### DIFF
--- a/src/TSMapEditor/Mutations/Classes/HeightMutations/LowerCellsMutation.cs
+++ b/src/TSMapEditor/Mutations/Classes/HeightMutations/LowerCellsMutation.cs
@@ -45,7 +45,7 @@ namespace TSMapEditor.Mutations.Classes.HeightMutations
             else
             {
                 var targetTile = MutationTarget.Map.GetTile(targetCellCoords);
-                var tilesToProcess = FillTerrainAreaMutation.GetFillAreaTiles(targetTile, MutationTarget.Map, MutationTarget.TheaterGraphics);
+                var tilesToProcess = Helpers.GetFillAreaTiles(targetTile, MutationTarget.Map, MutationTarget.TheaterGraphics);
 
                 // Process tiles
                 foreach (Point2D cellCoords in tilesToProcess)
--- a/src/TSMapEditor/Mutations/Classes/HeightMutations/LowerCellsMutation.cs
+++ b/src/TSMapEditor/Mutations/Classes/HeightMutations/LowerCellsMutation.cs
@@ -10,33 +10,52 @@ namespace TSMapEditor.Mutations.Classes.HeightMutations
     /// </summary>
     public class LowerCellsMutation : Mutation
     {
-        public LowerCellsMutation(IMutationTarget mutationTarget, Point2D targetCellCoords, BrushSize brushSize) : base(mutationTarget)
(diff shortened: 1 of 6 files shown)


Rampastring (changes requested):
I fully support the idea of this.

However implementation-wise, I want the mutations to be as self-contained as possible.

Having `LowerCellsMutation` and `RaiseCellsMutation` depend on a mutation for an entirely different tool is unacceptable.

As far as I can see, the best way to solve this would be splitting `GetFillAreaTiles` into a helper function in another class, either in the existing `Helpers` or in its own file.

Another option would be making a base class mutation that supported gathering an area to fill and deriving `FillTerrainAreaMutation`, `LowerCellsMutation` and `RaiseCellsMutation` from it, but in this case I think that would be more complex and less flexible then simply having a static helper method.

Starkku:
Moved the helper function to `Helpers.cs`.

src/TSMapEditor/Mutations/Classes/HeightMutations/LowerCellsMutation.cs
```diff
@@ -10,33 +10,52 @@ namespace TSMapEditor.Mutations.Classes.HeightMutations
...
+            {
+                var targetTile = MutationTarget.Map.GetTile(targetCellCoords);
+                var tilesToProcess = Helpers.GetFillAreaTiles(targetTile, MutationTarget.Map, MutationTarget.TheaterGraphics);
+
+                // Process tiles
+                foreach (Point2D cellCoords in tilesToProcess)
                 {
+                    var cell = MutationTarget.Map.GetTile(cellCoords);
+
                     cell.Level--;
```
Rampastring, line 59: As far as I can see, there is nothing preventing this from being executed when the level is already at 0?

src/TSMapEditor/Mutations/Classes/HeightMutations/RaiseCellsMutation.cs
```diff
@@ -10,33 +11,52 @@ namespace TSMapEditor.Mutations.Classes.HeightMutations
...
+            {
+                var targetTile = MutationTarget.Map.GetTile(targetCellCoords);
+                var tilesToProcess = Helpers.GetFillAreaTiles(targetTile, MutationTarget.Map, MutationTarget.TheaterGraphics);
+
+                // Process tiles
+                foreach (Point2D cellCoords in tilesToProcess)
                 {
+                    var cell = MutationTarget.Map.GetTile(cellCoords);
+
                     cell.Level++;
```
Rampastring, line 60: As far as I can see, there is nothing preventing this from being executed when the level is already at `Constants.MaxMapHeightLevel`?

Rampastring (changes requested):

Starkku:
Added the missing height checks.

Rampastring (approved):

PR #180, opened 2024-06-14
Adjust helper functions to convert waypoints longer than 2 letters

ZivDero:
Adjusts the converted functions to handle waypoints past ZZ. Credits to (I think) secsome for the original implementation in Phobos.

### DIFF
--- a/src/TSMapEditor/Helpers.cs
+++ b/src/TSMapEditor/Helpers.cs
@@ -180,7 +180,7 @@ namespace TSMapEditor
                 int c = str[i];
 
                 if (c is < 'A' or > 'Z')
-                    throw new InvalidOperationException("Waypoints may only contains characters A through Z: " + str); ;
+                    throw new InvalidOperationException("Waypoints may only contain characters A through Z: " + str); ;
 
                 n += (c - 64) * j;
             }
--- a/src/TSMapEditor/Helpers.cs
+++ b/src/TSMapEditor/Helpers.cs
@@ -185,7 +185,7 @@ namespace TSMapEditor
                 if (c is < 'A' or > 'Z')
                     throw new InvalidOperationException("Waypoints may only contain characters A through Z, invalid input: " + str); ;
 
-                n += (c - 64) * j;
+                n += (c - '@') * j; // '@' = 'A' - 1
             }
 
             return (n - 1);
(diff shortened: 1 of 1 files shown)


src/TSMapEditor/Helpers.cs
```diff
@@ -170,41 +171,40 @@ public static bool IsLandTypeWater(int landType)
 
         public static int GetWaypointNumberFromAlphabeticalString(string str)
         {
-            if (string.IsNullOrEmpty(str))
-                return -1;
+            const int charCount = 26;
+            str = str.ToUpperInvariant();
```
Starkku: Removing the null check here causes the editor to crash when trying to create new TeamType in the TeamTypes window, as the waypoint string on fresh TeamTypes defaults to null.

ZivDero: Good point, fixed

src/TSMapEditor/Helpers.cs
```diff
@@ -173,38 +174,40 @@ public static int GetWaypointNumberFromAlphabeticalString(string str)
...
+            int n = 0;
+            for (int i = str.Length - 1, j = 1; i >= 0; i--, j *= charCount)
+            {
+                int c = str[i];
+
+                if (c is < 'A' or > 'Z')
+                    throw new InvalidOperationException("Waypoints may only contain characters A through Z, invalid input: " + str); ;
 
-            const int CharCount = 26;
+                n += (c - 64) * j;
```
Rampastring: Put this `64` in a named constant. Yeah, we can figure out it's the char code of 'A' reduced by one, but it should be made obvious while reading the code.

src/TSMapEditor/Helpers.cs
```diff
@@ -173,38 +174,40 @@ public static int GetWaypointNumberFromAlphabeticalString(string str)
...
-            int secondLetterValue = waypointNumber % CharCount;
-
-            if (firstLetterValue == 0)
-                return ((char)('A' + secondLetterValue)).ToString();
+            while (waypointNumber > 0)
+            {
+                int m = waypointNumber % 26;
+                if (m == 0) m = 26;
+                sb.Insert(0, (char)(m + '@')); // '@' = 'A' - 1
+                waypointNumber = (waypointNumber - m) / 26;
```
Rampastring: Should reintroduce CharCount for this magic constant of 26.

Rampastring (changes requested):

Rampastring (approved):

PR #190, opened 2024-09-07
Filter, sort and color items in Triggers, Scripts and Team Types windows

JoyfulShush:
Adds the following QoL for WAE:

* Triggers:
    * Can now be filtered by entering parts of the trigger's name.
    * Now has new "Clone Event" and "Clone Action" buttons above the events and actions screen (no new logic, allows easy access to the existing context menu item with the same name.

* Scripts:
    * Can now be filtered by entering parts of the script's name.
    * Individual scripts can now have colored text.
    * Can now be sorted based on INI Name (ID), name, color, or color then name (same as with Triggers sort).

* Team Types:
    * Can now be filtered by entering parts of the Team Type's name.
    * Team Types can now override their text color (which defaults to the house's color) with a different color, which takes precedence.
    * Can now be sorted based on INI Name (ID), name, color, or color then name (same as with Triggers sort)
    * Bug fix: fixes an issue where deleting the last Team Type and creating a new one makes the selection not apply properly.

### DIFF
--- a/src/TSMapEditor/Models/Script.cs
+++ b/src/TSMapEditor/Models/Script.cs
@@ -78,7 +78,7 @@ namespace TSMapEditor.Models
             }
         }
 
-        public static NamedColor[] SupportedColors = NamedColors.GenericSupportedNamedColors;
+        public static NamedColor[] SupportedColors => NamedColors.GenericSupportedNamedColors;
 
         public Color XNAColor;
 
@@ -113,17 +113,22 @@ namespace TSMapEditor.Models
             return script;
         }
 
-        public void WriteToIniSection(IniFile iniFile, IniSection scriptSection)
+        public void WriteToIniSection(IniSection scriptSection)
         {
             for (int i = 0; i < Actions.Count; i++)
             {
                 scriptSection.SetStringValue(i.ToString(), $"{Actions[i].Action},{Actions[i].Argument}");
             }
 
-            scriptSection.SetStringValue("Name", Name);
+            scriptSection.SetStringValue("Name", Name);            
+        }
 
-            if (EditorColor != null)
-                iniFile.SetStringValue("EditorScriptInfo", ININame, EditorColor);
+        public void WriteEditorProperties(IniFile iniFile)
+        {
+            if (EditorColor == null)
+                return;
+
+            iniFile.SetStringValue("EditorScriptInfo", ININame, EditorColor);
         }
 
         public static Script ParseScript(string id, IniSection scriptSection)
--- a/src/TSMapEditor/Models/Script.cs
+++ b/src/TSMapEditor/Models/Script.cs
@@ -1,5 +1,9 @@
 ﻿using Rampastring.Tools;
 using System.Collections.Generic;
+using static TSMapEditor.Misc.NamedColors;
+using TSMapEditor.Misc;
+using System;
+using Microsoft.Xna.Framework;
 
 namespace TSMapEditor.Models
 {
@@ -46,6 +50,38 @@ namespace TSMapEditor.Models
             ININame = iniName;
         }
 
+        private string _editorColor;
+        /// <summary>
+        /// Editor-only. The color of the script in the UI.
+        /// If null, the script should be displayed with the default UI text color.
+        /// </summary>
+        public string EditorColor
+        {
+            get => _editorColor;
+            set
+            {
+                _editorColor = value;
+
+                if (_editorColor != null)
+                {
+                    int index = Array.FindIndex(SupportedColors, c => c.Name == value);
+                    if (index > -1)
+                    {
+                        XNAColor = SupportedColors[index].Value;
+                    }
+                    else
+                    {
+                        // Only allow assigning colors that actually exist in the color table
+                        _editorColor = null;
+                    }
+                }
+            }
+        }
+
+        public static NamedColor[] SupportedColors = NamedColors.GenericSupportedNamedColors;
+
+        public Color XNAColor;
+
         public string GetInternalID() => ININame;
         public void SetInternalID(string id) => ININame = id;
 
@@ -77,7 +113,7 @@ namespace TSMapEditor.Models
             return script;
         }
 
-        public void WriteToIniSection(IniSection scriptSection)
+        public void WriteToIniSection(IniFile iniFile, IniSection scriptSection)
         {
             for (int i = 0; i < Actions.Count; i++)
             {
@@ -85,6 +121,9 @@ namespace TSMapEditor.Models
             }
 
             scriptSection.SetStringValue("Name", Name);
+
+            if (EditorColor != null)
+                iniFile.SetStringValue("EditorScriptInfo", ININame, EditorColor);
         }
 
         public static Script ParseScript(string id, IniSection scriptSection)
--- a/src/TSMapEditor/Models/Script.cs
+++ b/src/TSMapEditor/Models/Script.cs
@@ -1,6 +1,5 @@
 ﻿using Rampastring.Tools;
 using System.Collections.Generic;
-using static TSMapEditor.Misc.NamedColors;
 using TSMapEditor.Misc;
 using System;
 using Microsoft.Xna.Framework;
--- a/src/TSMapEditor/Models/TeamType.cs
+++ b/src/TSMapEditor/Models/TeamType.cs
@@ -3,6 +3,8 @@ using Rampastring.Tools;
 using System;
 using System.Collections.Generic;
 using System.Text;
+using static TSMapEditor.Misc.NamedColors;
+using TSMapEditor.Misc;
 
 namespace TSMapEditor.Models
 {
@@ -135,7 +137,7 @@ namespace TSMapEditor.Models
             return clone;
         }
 
-        public void WriteToIniSection(IniSection iniSection, List<TeamTypeFlag> teamTypeFlags)
+        public void WriteToIniSection(IniFile iniFile, IniSection iniSection, List<TeamTypeFlag> teamTypeFlags)
         {
             // This cuts it for all properties of standard types
             WritePropertiesToIniSection(iniSection);
@@ -158,6 +160,9 @@ namespace TSMapEditor.Models
             {
                 iniSection.SetBooleanValue(flag.Name, IsFlagEnabled(flag.Name), BooleanStringStyle.YESNO_LOWERCASE);
             });
+
+            if (EditorColor != null)
+                iniFile.SetStringValue("EditorTeamTypeInfo", ININame, EditorColor);
         }
 
         public override RTTIType WhatAmI()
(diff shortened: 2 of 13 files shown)


src/TSMapEditor/Models/Script.cs
```diff
@@ -46,6 +50,38 @@ public Script(string iniName)
...
+                    else
+                    {
+                        // Only allow assigning colors that actually exist in the color table
+                        _editorColor = null;
+                    }
+                }
+            }
+        }
+
+        public static NamedColor[] SupportedColors = NamedColors.GenericSupportedNamedColors;
```
Rampastring: Make this a getter instead of a global variable.

```suggestion
        public static NamedColor[] SupportedColors => NamedColors.GenericSupportedNamedColors;
```

src/TSMapEditor/Models/Script.cs
```diff
@@ -77,14 +113,17 @@ public Script Clone(string iniName)
             return script;
         }
 
-        public void WriteToIniSection(IniSection scriptSection)
+        public void WriteToIniSection(IniFile iniFile, IniSection scriptSection)
```
Rampastring: This method no longer does what its name says if you add it an IniFile parameter.

It'd be cleaner if you didn't modify this function, but added a `WriteEditorProperties(IniFile iniFile)` method instead, and called it from `MapWriter`. That way saving editor-only properties and actual map script data can be kept separate, which is good for scripting purposes.

src/TSMapEditor/Models/TeamType.cs
```diff
@@ -158,13 +160,48 @@ public void WriteToIniSection(IniSection iniSection, List<TeamTypeFlag> teamType
...
+                iniFile.SetStringValue("EditorTeamTypeInfo", ININame, EditorColor);
         }
 
         public override RTTIType WhatAmI()
         {
             return RTTIType.TeamType;
         }
 
-        public Color GetXNAColor() => Helpers.GetHouseTypeUITextColor(HouseType);
+        public static NamedColor[] SupportedColors = NamedColors.GenericSupportedNamedColors;
```
Rampastring: Make this a getter instead of a global variable.

```suggestion
        public static NamedColor[] SupportedColors => NamedColors.GenericSupportedNamedColors;
```

src/TSMapEditor/Models/TeamType.cs
```diff
@@ -158,13 +160,48 @@ public void WriteToIniSection(IniSection iniSection, List<TeamTypeFlag> teamType
...
+                    }
+                    else
+                    {
+                        // Only allow assigning colors that actually exist in the color table
+                        _editorColor = null;
+                    }
+                }
+            }
+        }
+        private Color EditorColorValue;
```
Rampastring: Needs spacing between end of block and field.

```suggestion
        }

        private Color EditorColorValue;
```

src/TSMapEditor/Models/Trigger.cs
```diff
@@ -2,45 +2,18 @@
 using Rampastring.Tools;
 using System;
 using System.Collections.Generic;
+using TSMapEditor.Misc;
 using TSMapEditor.Models.Enums;
+using static TSMapEditor.Misc.NamedColors;
```
Rampastring: Is this `using` clause needed?

src/TSMapEditor/Models/Waypoint.cs
```diff
@@ -3,6 +3,7 @@
 using Microsoft.Xna.Framework;
 using System;
 using System.Globalization;
+using static TSMapEditor.Misc.NamedColors;
```
Rampastring: What is this for? It's the only change in this file.

You could refactor this file to also use the generic colors instead of having its own array.

src/TSMapEditor/UI/Windows/ScriptsWindow.cs
```diff
@@ -515,15 +581,49 @@ private void ListScripts()
         {
             lbScriptTypes.Clear();
 
-            foreach (var script in map.Scripts)
+            List<Script> sortedScripts = map.Scripts;
```
Rampastring: There's a few things that could be improved here. I'll be adding a few suggestions, they make sense when viewed as a whole.

```suggestion
            IEnumerable<Script> sortedScripts = map.Scripts;
```

src/TSMapEditor/UI/Windows/ScriptsWindow.cs
```diff
@@ -515,15 +581,49 @@ private void ListScripts()
...
+
+            var shouldViewTop = false; // when filtering the scroll bar should update so we use a flag here
+            if (tbFilter.Text != string.Empty && tbFilter.Text != tbFilter.Suggestion)
             {
-                lbScriptTypes.AddItem(new XNAListBoxItem() { Text = script.Name, Tag = script });
+                sortedScripts = sortedScripts.Where(sortedScript => sortedScript.Name.Contains(tbFilter.Text, StringComparison.CurrentCultureIgnoreCase)).ToList();
```
Rampastring: No need to turn this into a list since we'll still use Linq to work on it during the sort. You'll just cause an extra memory allocation and heavier syntax. Also, there is likely no need to turn this into a list at all since IIRC `IEnumerable` can be worked with using `foreach`.

Further, you're calling the variable inside the lambda `sortedScript`, which makes no sense. As a singular script instance, it's just a script.

```suggestion
                sortedScripts = sortedScripts.Where(script => script.Name.Contains(tbFilter.Text, StringComparison.CurrentCultureIgnoreCase));
```

src/TSMapEditor/UI/Windows/ScriptsWindow.cs
```diff
@@ -515,15 +581,49 @@ private void ListScripts()
...
+                case ScriptSortMode.Name:
+                    sortedScripts = sortedScripts.OrderBy(script => script.Name).ThenBy(script => script.ININame).ToList();
+                    break;
+                case ScriptSortMode.ColorThenName:
+                    sortedScripts = sortedScripts.OrderBy(script => script.EditorColor).ThenBy(script => script.Name).ToList();
+                    break;
+                case ScriptSortMode.ID:
+                default:
+                    sortedScripts = sortedScripts.OrderBy(script => script.ININame).ToList();
+                    break;
```
Rampastring: No need to turn it into a list.

```suggestion
                case ScriptSortMode.Color:
                    sortedScripts = sortedScripts.OrderBy(script => script.EditorColor).ThenBy(script => script.ININame);
                    break;
                case ScriptSortMode.Name:
                    sortedScripts = sortedScripts.OrderBy(script => script.Name).ThenBy(script => script.ININame);
                    break;
                case ScriptSortMode.ColorThenName:
                    sortedScripts = sortedScripts.OrderBy(script => script.EditorColor).ThenBy(script => script.Name);
                    break;
                case ScriptSortMode.ID:
                default:
                    sortedScripts = sortedScripts.OrderBy(script => script.ININame);
                    break;
```

src/TSMapEditor/UI/Windows/ScriptsWindow.cs
```diff
@@ -551,7 +652,12 @@ private void EditScript(Script script)
                 });
             }
 
+            ddScriptColor.SelectedIndex = ddScriptColor.Items.FindIndex(item => item.Text == editedScript.EditorColor);
+            if (ddScriptColor.SelectedIndex == -1)
```
Rampastring: Just a personal preference, to be cleaner with the style of the rest of the codebase.

```suggestion
            if (ddScriptColor.SelectedIndex < 0)
```

src/TSMapEditor/Models/TeamType.cs
```diff
@@ -135,7 +137,7 @@ public TeamType Clone(string iniName)
             return clone;
         }
 
-        public void WriteToIniSection(IniSection iniSection, List<TeamTypeFlag> teamTypeFlags)
+        public void WriteToIniSection(IniFile iniFile, IniSection iniSection, List<TeamTypeFlag> teamTypeFlags)
```
Rampastring: Same thing here as with the scripts comment above.

src/TSMapEditor/UI/Windows/TeamTypesWindow.cs
```diff
@@ -308,11 +362,45 @@ private void ListHouses()
...
+                    sortedTeamTypes = sortedTeamTypes.OrderBy(teamType => teamType.Name).ThenBy(teamType => teamType.ININame).ToList();
+                    break;
+                case TeamTypeSortMode.ColorThenName:
+                    sortedTeamTypes = sortedTeamTypes.OrderBy(teamType => teamType.GetXNAColor().ToString()).ThenBy(teamType => teamType.Name).ToList();
+                    break;
+                case TeamTypeSortMode.ID:
+                default:
+                    sortedTeamTypes = sortedTeamTypes.OrderBy(teamType => teamType.ININame).ToList();
+                    break;
+            }
```
Rampastring: All the teamtype-iterating code here can be improved similarly to the implementation in the scripts window.

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1417,6 +1423,14 @@ private void BtnDeleteAction_LeftClick(object sender, EventArgs e)
             EditTrigger(editedTrigger);
         }
 
+        private void BtnCloneAction_LeftClick(object sender, EventArgs e)
+        {
+            if (editedTrigger == null || lbActions.SelectedItem == null)
+                return;
+
+            CloneEventOrAction(lbActions, editedTrigger.Actions);
+        }
```
Rampastring: `CloneEventOrAction` already does verification on active trigger and list box selected item.

```suggestion
        private void BtnCloneAction_LeftClick(object sender, EventArgs e) => CloneEventOrAction(lbActions, editedTrigger.Actions);
```

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1435,6 +1449,14 @@ private void BtnDeleteEvent_LeftClick(object sender, EventArgs e)
             EditTrigger(editedTrigger);
         }
 
+        private void BtnCloneEvent_LeftClick(object sender, EventArgs e)
+        {
+            if (editedTrigger == null || lbEvents.SelectedItem == null)
+                return;
+
+            CloneEventOrAction(lbEvents, editedTrigger.Conditions);
+        }
```
Rampastring:
```suggestion
        private void BtnCloneEvent_LeftClick(object sender, EventArgs e) => CloneEventOrAction(lbEvents, editedTrigger.Conditions);
```

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1464,6 +1486,14 @@ private void ListTriggers()
             lbTriggers.Clear();
 
             List<Trigger> sortedTriggers = map.Triggers;
```
Rampastring: Could also turn this into `IEnumerable<Trigger>` and get rid of the `ToList` calls.

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -1647,6 +1680,11 @@ private void TbName_TextChanged(object sender, EventArgs e)
             lbTriggers.SelectedItem.Text = tbName.Text;
         }
 
+        private void TbFilter_TextChanged(object sender, EventArgs e)
+        {
+            ListTriggers();
+        }
```
Rampastring:
```suggestion
        private void TbFilter_TextChanged(object sender, EventArgs e) => ListTriggers();
```

Rampastring (changes requested):

Rampastring (approved):

PR #197, opened 2024-09-21
Transform AI Trigger Object Type field into a popup picker

JoyfulShush:
Currently, if a user wants to make an AI Trigger that has a condition which relies on an object (e.g. "enemy has more or equal to 3 AGTs"), the user must write the ININame of the object manually in a text box, which is inconvenient, non-human readable and error prone.

This PR aims to transform the text box into a popup picker, which is already used in many other places such as Triggers and Team Types. This allows the user to utilize the powerful popup picker's capabilities, such as listing all units and structures, picking them from a list, and searching them from a list.

Since most of the logic for this selection window already existed, only minor changes were needed to apply it.
Includes handling of loading and saving properly into files.

Example screenshot:
image

### DIFF
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -995,7 +995,7 @@ namespace TSMapEditor.Initialization
                         AddMapLoadError($"AITriggerType {kvp.Key} has a non-existent condition object \"{parts[5]}\"");
                     }
 
-                    aiTriggerType.ConditionObject = map.Rules.FindTechnoType(parts[5]);
+                    aiTriggerType.ConditionObject = conditionObject;
                 }
 
                 aiTriggerType.LoadedComparatorString = parts[6];
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -995,7 +995,7 @@ namespace TSMapEditor.Initialization
                         AddMapLoadError($"AITriggerType {kvp.Key} has a non-existent condition object \"{parts[5]}\"");
                     }
 
-                    aiTriggerType.ConditionObjectString = parts[5];
+                    aiTriggerType.ConditionObject = map.Rules.FindTechnoType(parts[5]);
                 }
 
                 aiTriggerType.LoadedComparatorString = parts[6];
--- a/src/TSMapEditor/Models/Map.cs
+++ b/src/TSMapEditor/Models/Map.cs
@@ -1833,16 +1833,6 @@ namespace TSMapEditor.Models
                 }
             }
 
-            // Check for AITriggerTypes that have a non-existing condition object
-            foreach (var aiTrigger in AITriggerTypes)
-            {
-                if (aiTrigger.ConditionObject != null &&
-                    Rules.FindTechnoType(aiTrigger.ConditionObject.ININame) == null)
-                {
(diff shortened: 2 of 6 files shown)


src/TSMapEditor/Models/Map.cs
```diff
@@ -1836,11 +1836,10 @@ public List<string> CheckForIssues()
...
             foreach (var aiTrigger in AITriggerTypes)
             {
-                if (!string.IsNullOrWhiteSpace(aiTrigger.ConditionObjectString) &&
-                    !Helpers.IsStringNoneValue(aiTrigger.ConditionObjectString) &&
-                    Rules.FindTechnoType(aiTrigger.ConditionObjectString) == null)
+                if (aiTrigger.ConditionObject != null &&
+                    Rules.FindTechnoType(aiTrigger.ConditionObject.ININame) == null)
                 {
-                    issueList.Add($"AITrigger '{aiTrigger.Name}' has a condition object '{aiTrigger.ConditionObjectString}' that does not exist in Rules!");
+                    issueList.Add($"AITrigger '{aiTrigger.Name}' has a condition object '{aiTrigger.ConditionObject.GetEditorDisplayName()} ({aiTrigger.ConditionObject.ININame})' that does not exist in Rules!");
```
Rampastring: If the condition object is non-null, then it already exists in Rules by definition. This check can be removed.

src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -995,7 +995,7 @@ public static void ReadAITriggerTypes(IMap map, IniFile mapIni)
                         AddMapLoadError($"AITriggerType {kvp.Key} has a non-existent condition object \"{parts[5]}\"");
                     }
 
-                    aiTriggerType.ConditionObjectString = parts[5];
+                    aiTriggerType.ConditionObject = map.Rules.FindTechnoType(parts[5]);
```
Rampastring: No need to look up the object again when it's already done 7 lines earlier :)

```suggestion
                    aiTriggerType.ConditionObject = conditionObject;
```

src/TSMapEditor/UI/Windows/AITriggersWindow.cs
```diff
@@ -113,7 +119,7 @@ private void BtnNew_LeftClick(object sender, EventArgs e)
             var aiTrigger = new AITriggerType(map.GetNewUniqueInternalId());
             aiTrigger.Name = "New AITrigger";
             aiTrigger.OwnerName = "<all>";
-            aiTrigger.ConditionObjectString = string.Empty;
+            aiTrigger.ConditionObject = null;
```
Rampastring: This line can be removed. Object references are null by default.

src/TSMapEditor/UI/Windows/SelectTechnoTypeWindow.cs
```diff
@@ -35,6 +37,9 @@ protected override void ListObjects()
         {
             lbObjectList.Clear();
 
+            if (IncludeNone)
+                lbObjectList.AddItem("None");
```
Rampastring, line 41: Did you test to make sure that selecting this works and has the desired effect on the AITrigger? Just making sure.

Rampastring (changes requested):

src/TSMapEditor/UI/Windows/SelectTechnoTypeWindow.cs
chkp-osheral: Yes, indeed. It ends up with an empty string and a null tag on the condition object.

JoyfulShush: Yes, works as intended. Line becomes empty, ConditionObject is set to null, and when saving - saves to <none>.

Rampastring (approved):

PR #198, opened 2024-09-26
Handling Individual Infantry in Cells

JoyfulShush:
Not to be confused with #199 , which allows handling entire infantry groups when holding a button.

This PR aims to address logic for handling individual infantry that reside in the same cell. Currently, WAE makes it impossible for the user to select and handle a specific infantry out of multiple that can reside in a cell; instead, it will always default to whichever infantry occupies the "first" cell. This means that moving specific infantry, rotating them, or assigning missions to them is not possible without isolating them in a different cell, then bringing them back.

The logic in this PR is changed to instead pass a relative position in a tile based on the cursor's position. This allows a specific cell that houses infantry to look at its sub cells and figure out which is closest to the cursor, allowing you to handle a specific infantry that is as close to the mouse as possible.

Also adjusts line draws slightly to address this change, showing the infantry that is being moved.

Showcase gif:

https://github.com/user-attachments/assets/e3bea3af-df7c-4720-8ddd-64a401cfd142

### DIFF
--- a/src/TSMapEditor/Constants.cs
+++ b/src/TSMapEditor/Constants.cs
@@ -1,5 +1,6 @@
 ﻿using Rampastring.Tools;
 using System;
+using TSMapEditor.GameMath;
 
 namespace TSMapEditor
 {
@@ -107,6 +108,11 @@ namespace TSMapEditor
 
         public const string VeinholeMonsterTypeName = "VEINHOLE";
         public const string VeinholeDummyTypeName = "VEINHOLEDUMMY";
+
+        public static Point2D SubCellTopOffSet = new Point2D(0, CellSizeY / -4);
+        public static Point2D SubCellBottomOffSet = new Point2D(0, CellSizeY / 4);
+        public static Point2D SubCellLeftOffSet = new Point2D(CellSizeX / -4, 0);
+        public static Point2D SubCellRightOffSet = new Point2D(CellSizeX / 4, 0);
 
         public static void Init()
         {
--- a/src/TSMapEditor/Constants.cs
+++ b/src/TSMapEditor/Constants.cs
@@ -107,12 +107,7 @@ namespace TSMapEditor
         public const char NewTheaterGenericLetter = 'G';
 
         public const string VeinholeMonsterTypeName = "VEINHOLE";
-        public const string VeinholeDummyTypeName = "VEINHOLEDUMMY";
-
-        public static Point2D SubCellTopOffSet = new Point2D(0, CellSizeY / -4);
-        public static Point2D SubCellBottomOffSet = new Point2D(0, CellSizeY / 4);
-        public static Point2D SubCellLeftOffSet = new Point2D(CellSizeX / -4, 0);
-        public static Point2D SubCellRightOffSet = new Point2D(CellSizeX / 4, 0);
+        public const string VeinholeDummyTypeName = "VEINHOLEDUMMY";        
 
         public static void Init()
         {
--- a/src/TSMapEditor/Constants.cs
+++ b/src/TSMapEditor/Constants.cs
@@ -107,7 +107,7 @@ namespace TSMapEditor
         public const char NewTheaterGenericLetter = 'G';
 
         public const string VeinholeMonsterTypeName = "VEINHOLE";
-        public const string VeinholeDummyTypeName = "VEINHOLEDUMMY";        
+        public const string VeinholeDummyTypeName = "VEINHOLEDUMMY";
 
         public static void Init()
         {
--- a/src/TSMapEditor/Models/MapTile.cs
+++ b/src/TSMapEditor/Models/MapTile.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Xna.Framework;
 using TSMapEditor.GameMath;
 using TSMapEditor.Models.Enums;
 using TSMapEditor.Models.MapFormat;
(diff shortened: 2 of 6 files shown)


src/TSMapEditor/Models/MapTile.cs
```diff
@@ -356,5 +412,7 @@ public void ChangeTileIndex(int newTileIndex, byte newSubTileIndex)
         }
 
         public Point2D CoordsToPoint() => new Point2D(X, Y);
+
+        public Point2D GetTileCenter() => new Point2D(X + Constants.CellSizeX / 2, Y + Constants.CellSizeY / 2);
```
Rampastring: This is currently nonsensical as it's mixing map coords (X, Y) with pixel coords (Constants.CellSizeX / 2, Constants.CellSizeY / 2). I'd leave only the latter.

```suggestion
        public Point2D GetTileCenter() => new Point2D(Constants.CellSizeX / 2, Constants.CellSizeY / 2);
```

src/TSMapEditor/Models/MapTile.cs
```diff
@@ -198,6 +200,49 @@ public void DoForAllWaypoints(Action<Waypoint> action)
...
+                case SubCell.Left:
+                    return Constants.SubCellLeftOffSet;
+
+                case SubCell.Right:
+                    return Constants.SubCellRightOffSet;
+
+                default:
+                    return Point2D.Zero;
+            }
+        }
```
Rampastring: Move this to either `CellMath` or `Helpers` and make it static.

Then you can also get rid of the new constants, and have this function return the matching values directly.

`InfantryRenderer` can then be modified to call this function instead of containing the same block of duplicated code.

src/TSMapEditor/Rendering/MapView.cs
```diff
@@ -2021,5 +2041,14 @@ private void InstantRenderMinimap()
...
         }
+
+        private Point2D GetRelativeTilePositionFromCursorPosition(MapTile tile)
+        {
+            var tileCenter = tile.GetTileCenter();
+            var cursorPosition = GetCursorMapPoint() + new Point2D(0, Constants.MapYBaseline);
+            var offset = cursorPosition - CellMath.CellCenterPointFromCellCoords(tile.CoordsToPoint(), Map);
+
+            return tileCenter + offset;
+        }
```
Rampastring: This can be simplified, just like `MapTile.GetTileCenter`. I believe the following should work.

```suggestion
        private Point2D GetRelativeTilePositionFromCursorPosition(MapTile tile)
        {
            return GetCursorMapPoint() - CellMath.CellTopLeftPointFromCellCoords_NoBaseline(tile.CoordsToPoint(), Map);
        }
```

However for TS and RA2/YR support you'll need to also check if `!Constants.IsFlatWorld && !EditorState.Is2DMode`, and if so, call `CellTopLeftPointFromCellCoords_3D_NoBaseline` instead.

Rampastring (changes requested):

src/TSMapEditor/Constants.cs
```diff
@@ -107,12 +107,7 @@ public static class Constants
         public const char NewTheaterGenericLetter = 'G';
 
         public const string VeinholeMonsterTypeName = "VEINHOLE";
-        public const string VeinholeDummyTypeName = "VEINHOLEDUMMY";
-
-        public static Point2D SubCellTopOffSet = new Point2D(0, CellSizeY / -4);
-        public static Point2D SubCellBottomOffSet = new Point2D(0, CellSizeY / 4);
-        public static Point2D SubCellLeftOffSet = new Point2D(CellSizeX / -4, 0);
-        public static Point2D SubCellRightOffSet = new Point2D(CellSizeX / 4, 0);
+        public const string VeinholeDummyTypeName = "VEINHOLEDUMMY";
```
Rampastring: You've somehow added an additional bunch of spaces at the end of this line.

src/TSMapEditor/Models/MapTile.cs
```diff
@@ -196,6 +198,31 @@ public void DoForAllWaypoints(Action<Waypoint> action)
...
+            foreach (var subCell in subCells)
+            {
+                var subCellCoords = GetTileCenter() + CellMath.GetSubCellOffset(subCell);
+                var distanceToSubCell = Vector2.Distance(position.ToXNAVector(), subCellCoords.ToXNAVector());
+                if (distanceToSubCell < shortestDistance)
+                {
+                    closestSubcell = subCell;
+                    shortestDistance = distanceToSubCell;
+                }
+            }
```
Rampastring: This currently has an issue that it does not process all of the sub cells, only the normally usable ones. Unless I'm mistaken, that means that you can't move infantry that have somehow (through editing the map with another tool than WAE) ended up on the unusual subcell spots.

Also I think the code becomes simpler if we filter the subcells inside the loop. As a bonus, that way we can also avoid a memory allocation for the `subCells` array. Which is not very important in the typical usecase, but could be if someone ever wanted to use this function for some batch processing.

```suggestion
            SubCell closestSubcell = SubCell.None;
            float shortestDistance = float.MaxValue;

            for (int i = 0; i < SubCellCount; i++)
            {
                SubCell subCell = (SubCell)i;
                if (onlyOccupiedCells && GetInfantryFromSubCellSpot(subCell) == null)
                    continue;

                var subCellCoords = GetTileCenter() + CellMath.GetSubCellOffset(subCell);
                var distanceToSubCell = Vector2.Distance(position.ToXNAVector(), subCellCoords.ToXNAVector());
                if (distanceToSubCell < shortestDistance)
                {
                    closestSubcell = subCell;
                    shortestDistance = distanceToSubCell;
                }
            }
```

Rampastring (changes requested):
Looks better now. A couple small changes still necessary.

Rampastring (approved):

PR #214, opened 2024-12-16
Target .NET 8 for SecondStageUpdater

SadPencil:
.NET 7 has been end of life since May 14, 2024. Still targeting .net 7 causes that even a user has .NET 8.0 installed, he/she will be asked for installing .NET 7. As WAE has been targeting .NET 8 for a long time, there should be no .NET 7 exclusive components.

The new `SecondStageUpdater.exe` file was compiled from <https://github.com/Rampastring/Rampastring.Updater/pull/2>. Clicked `Rampastring.Updater\SecondStageUpdater\build.bat` and got `Rampastring.Updater\SecondStageUpdater\bin\Release\net8.0-windows\win-x64\publish\SecondStageUpdater.exe`.


Rampastring:
Thanks for bringing this to my attention. The binary has been updated to target .NET 8 in https://github.com/CnCNet/WorldAlteringEditor/commit/9cb279221f16236790a6aee4a9dc5792879d5920

PR #218, opened 2025-02-07
Add support for directional alpha images

Starkku:
Ares allows alpha image shapes with multiple frames to display different frames depending on facing of the object. This PR implements this logic in WAE.

It is not currently behind any toggle other than the alpha image itself having multiple frames. If there's a need to add a toggle, I can do that.

### DIFF
--- a/src/TSMapEditor/Rendering/MapView.cs
+++ b/src/TSMapEditor/Rendering/MapView.cs
@@ -1635,7 +1635,7 @@ namespace TSMapEditor.Rendering
                     if (frameCount > 1)
                     {
                         if (alphaImagesToRender[i].OwnerObject is TechnoBase ownerTechno)
-                            frame = ownerTechno.Facing / (256 / frameCount);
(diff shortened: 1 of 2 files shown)


src/TSMapEditor/Rendering/MapView.cs
```diff
@@ -1625,10 +1625,20 @@ private void DrawWorld()
...
+                    if (frameCount <= 0)
                         continue;
 
-                    var alphaTexture = alphaShape.GetFrame(0);
+                    int frame = 0;
+
+                    if (frameCount > 1)
+                    {
+                        if (alphaImagesToRender[i].OwnerObject is TechnoBase ownerTechno)
+                            frame = ownerTechno.Facing / (256 / frameCount);
```
Rampastring: I'd like to explain this 256 magic number here. Maybe the following would work?

PR looks good otherwise.

```suggestion
                            frame = ownerTechno.Facing / ((Constants.FacingMax + 1) / frameCount);
```

Rampastring (approved):

Rampastring (approved):

Issue #226, opened 2025-04-13
[Feature Request] Make Celltags Moveable


krnyoshi:
Self explanatory. Can we have it where we can manually move cell tags? FinalSun also allows for this.

Rampastring:
Implemented in https://github.com/CnCNet/WorldAlteringEditor/commit/c0733956ac5573f903f4e92493779fa61d198230

PR #227, opened 2025-04-13
Support moving or cloning Cell Tags

JoyfulShush:
Implements #226 

Also allows cloning Cell Tags.

Additionally, includes a small refactor for cloning support - moving the logic into a reusable helper function that is reused when appropriate.

### DIFF
--- a/src/TSMapEditor/Helpers.cs
+++ b/src/TSMapEditor/Helpers.cs
@@ -16,7 +16,7 @@ namespace TSMapEditor
 {
     public static class Helpers
     {
-        public static readonly RTTIType[] SupportedCloneTypes = [RTTIType.Terrain, RTTIType.CellTag];
+        public static readonly RTTIType[] SupportedCloneTypes = [ RTTIType.Terrain, RTTIType.CellTag ];
 
         public static bool IsStringNoneValue(string str)
         {
--- a/src/TSMapEditor/Helpers.cs
+++ b/src/TSMapEditor/Helpers.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using TSMapEditor.GameMath;
 using TSMapEditor.Models;
 using TSMapEditor.Models.Enums;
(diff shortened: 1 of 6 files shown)


src/TSMapEditor/Helpers.cs
```diff
@@ -15,6 +16,8 @@ namespace TSMapEditor
 {
     public static class Helpers
     {
+        public static readonly RTTIType[] SupportedCloneTypes = [RTTIType.Terrain, RTTIType.CellTag];
```
Rampastring:
```suggestion
        public static readonly RTTIType[] SupportedCloneTypes = [ RTTIType.Terrain, RTTIType.CellTag ];
```

src/TSMapEditor/Models/Map.cs
```diff
@@ -1104,6 +1111,12 @@ public bool CanPlaceObjectAt(IMovable movable, Point2D newCoords, bool blocksSel
             }
 
             MapTile cell = GetTile(newCoords);
+
+            if (movable.WhatAmI() == RTTIType.CellTag)
+            {
+                return cell.CellTag == null;
+            }
```
Rampastring: I prefer to have the shorter/simpler cases at the top of the function. Please move this block and `MapTile cell = GetTile(newCoords);` between the waypoint and building handling code.

src/TSMapEditor/Mutations/Classes/CloneObjectMutation.cs
```diff
@@ -12,10 +12,10 @@ public class CloneObjectMutation : Mutation
     {
         public CloneObjectMutation(IMutationTarget mutationTarget, IMovable movable, Point2D clonePosition) : base(mutationTarget)
         {
-            this.objectToClone = (AbstractObject)movable;
-            if (!objectToClone.IsTechno() && objectToClone.WhatAmI() != RTTIType.Terrain)
-                throw new NotSupportedException(nameof(CloneObjectMutation) + " only supports cloning Technos and TerrainObjects!");
+            if (!Helpers.IsCloningSupported(movable))
+                throw new NotSupportedException(nameof(CloneObjectMutation) + " only supports cloning Technos, TerrainObjects and Cell Tags!");
```
Rampastring: Keeping naming consistent with `TerrainObjects`.

```suggestion
                throw new NotSupportedException(nameof(CloneObjectMutation) + " only supports cloning Technos, TerrainObjects and CellTags!");
```

Rampastring (changes requested):
Looks mostly good, only a few stylistic nitpicks to fix.

Rampastring (approved):

PR #232, opened 2025-05-04
Add Home Cell and Theme keys to Basic Options window

JoyfulShush:
Adds two new keys to the Basic Options window:

* Home Cell - used to specify which waypoint the mission starts the camera in. Number between 0-699.
* Theme - used to specify which music theme, if any, the mission starts in. Can be added and removed at will.

image

Additionally, changes the text in the Place Waypoints from `0-99` to `0-699`, as we now support 700 waypoints.

### DIFF
--- a/src/TSMapEditor/Config/UI/Windows/BasicSectionConfigWindow.ini
+++ b/src/TSMapEditor/Config/UI/Windows/BasicSectionConfigWindow.ini
@@ -11,23 +11,28 @@ $CC06=tbPercent:EditorNumberTextBox
 $CC07=lblPercent:XNALabel
 $CC08=tbInitialTime:EditorNumberTextBox
 $CC09=lblInitialTime:XNALabel
-$CC10=chkEndOfGame:XNACheckBox
-$CC11=chkOneTimeOnly:XNACheckBox
-$CC12=chkSkipScore:XNACheckBox
-$CC13=chkSkipMapSelect:XNACheckBox
-$CC14=chkIgnoreGlobalAITriggers:XNACheckBox
+$CC10=tbHomeCell:EditorNumberTextBox
+$CC11=lblHomeCell:XNALabel
+$CC12=tbTheme:EditorTextBox
+$CC13=btnThemePreset:EditorButton
+$CC14=lblTheme:XNALabel
+$CC15=chkEndOfGame:XNACheckBox
+$CC16=chkOneTimeOnly:XNACheckBox
+$CC17=chkSkipScore:XNACheckBox
+$CC18=chkSkipMapSelect:XNACheckBox
 ; Second column
-$CC15=chkOfficial:XNACheckBox
-$CC16=chkTruckCrate:XNACheckBox
-$CC17=chkTrainCrate:XNACheckBox
-$CC18=chkMultiplayerOnly:XNACheckBox
-$CC19=chkGrowingTiberium:XNACheckBox
-$CC20=chkGrowingVeins:XNACheckBox
-$CC21=chkGrowingIce:XNACheckBox
-$CC22=chkTiberiumDeathToVisceroid:XNACheckBox
-$CC23=chkFreeRadar:XNACheckBox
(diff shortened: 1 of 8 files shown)


src/TSMapEditor/Config/UI/Windows/PlaceWaypointWindow.ini
```diff
@@ -11,7 +11,7 @@ HasCloseButton=yes
 [lblDescription]
 $X=EMPTY_SPACE_SIDES
 $Y=EMPTY_SPACE_TOP
-Text=Input waypoint number (0-99):
+Text=Input waypoint number (0-699):
```
Rampastring: Please instead modify `PlaceWaypointWindow.cs` to modify the string dynamically depending on the value of `Constants.MaxWaypoint`, different games can have different waypoint limits so a hardcoded value here will always be incorrect for some of them.

Rampastring (changes requested):

src/TSMapEditor/Config/UI/Windows/PlaceWaypointWindow.ini
JoyfulShush: Per our offline discussion, we decided to hardcode the width of the window to 215 and have the label be dynamically assigned. All other elements are based off the hardcoded window width.

src/TSMapEditor/Config/UI/Windows/BasicSectionConfigWindow.ini
```diff
@@ -11,23 +11,28 @@ $CC06=tbPercent:EditorNumberTextBox
...
 $CC08=tbInitialTime:EditorNumberTextBox
 $CC09=lblInitialTime:XNALabel
-$CC10=chkEndOfGame:XNACheckBox
-$CC11=chkOneTimeOnly:XNACheckBox
-$CC12=chkSkipScore:XNACheckBox
-$CC13=chkSkipMapSelect:XNACheckBox
-$CC14=chkIgnoreGlobalAITriggers:XNACheckBox
+$CC10=tbHomeCell:EditorNumberTextBox
+$CC11=lblHomeCell:XNALabel
+$CC12=tbTheme:EditorTextBox
```
Rampastring: I wouldn't use a text box here - it is challenging for users to type a full song name by themselves (and apparently your current code does not even make that possible, giving users a false impression of functionality with the text box).

I think this is a natural fit for an `EditorPopUpSelector` control instead. For example, see how the TaskForce and Script selectors in `TeamTypesWindow` work.

With it, you can also get rid of `btnThemePreset` as just clicking on the pop-up selector should open up the theme selection window.

Rampastring (changes requested):

src/TSMapEditor/UI/Windows/BasicSectionConfigWindow.cs
```diff
@@ -37,7 +41,9 @@ public BasicSectionConfigWindow(WindowManager windowManager, Map map) : base(win
         private XNACheckBox chkTiberiumDeathToVisceroid;
         private XNACheckBox chkFreeRadar;
         private XNACheckBox chkRequiredAddOn;
+        private EditorButton btnApply;
```
Rampastring: As far as I can see, there is no reason for us to define `btnApply` as a field as it is only used in `Initialize`, where it was previously fetched and used. It has no reason to exist elsewhere. Was there a reason why you added it here?

Rampastring (changes requested):
Looks good aside from defining `btnApply` as a field.

src/TSMapEditor/UI/Windows/BasicSectionConfigWindow.cs
chkp-osheral: Guess not. I'll remove it. I was trying to bring it inline with the rest of the fields, but now I understand why it wasn't one in the first place.

Rampastring (approved):

PR #241, opened 2025-06-01
Implement support for saving OverlayPack as shorts with NewINIFormat == 5

ZivDero:


src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -693,7 +695,19 @@ public static void ReadOverlays(IMap map, IniFile mapIni)
                         continue;
 
                     int overlayDataIndex = (tile.Y * Constants.MAX_MAP_LENGTH_IN_DIMENSION) + tile.X;
-                    int overlayTypeIndex = uncompressedOverlayPack[overlayDataIndex];
+
+                    int overlayTypeIndex;
+                    if (needsExtendedOverlayPack)
+                    {
+                        ushort index = (ushort)(uncompressedOverlayPack[overlayDataIndex * 2] | (uncompressedOverlayPack[overlayDataIndex * 2 + 1] << 8));
```
Rampastring: I'd like to use `BinaryPrimitives.ReadInt16LittleEndian` or `BinaryPrimitives.ReadInt16BigEndian` instead of doing this manually, but that would require converting the byte array into a span.

src/TSMapEditor/Models/Map.cs
```diff
@@ -357,6 +357,21 @@ private void Write(string filePath = null)
...
+            // Determine if we need to save as NewINIFormat == 5
+            // If any of the overlays on the map have a heap ID > 255,
+            // then we have to use shorts to save OverlayPack
+            bool needsExtendedOverlayPack = false;
+            DoForAllValidTiles(tile =>
+            {
+                if (tile.Overlay?.OverlayType == null)
+                    return;
+
+                if (tile.Overlay.OverlayType.Index > 255)
```
Rampastring:
```suggestion
                if (tile.Overlay.OverlayType.Index > byte.MaxValue)
```

src/TSMapEditor/Initialization/MapWriter.cs
```diff
@@ -117,9 +117,11 @@ public static void WriteOverlays(IMap map, IniFile mapIni)
             mapIni.RemoveSection(overlayPackSectionName);
             mapIni.RemoveSection(overlayDataPackSectionName);
 
-            var overlayArray = new byte[Constants.MAX_MAP_LENGTH_IN_DIMENSION * Constants.MAX_MAP_LENGTH_IN_DIMENSION];
+            bool needsExtendedOverlayPack = map.Basic.NewINIFormat >= 5;
+
+            var overlayArray = new byte[Constants.MAX_MAP_LENGTH_IN_DIMENSION * Constants.MAX_MAP_LENGTH_IN_DIMENSION * (needsExtendedOverlayPack ? 2 : 1)];
             for (int i = 0; i < overlayArray.Length; i++)
-                overlayArray[i] = Constants.NO_OVERLAY;
+                overlayArray[i] = 0xFF; // fill the entire array with 0xFF, which will be (char)-1 or (short)-1, regardless of what format we're writing in
```
Rampastring: This comment seems to be incorrect. `char` is 2 bytes wide in C#. It also appears to be unsigned, meaning there is no such thing as `(char)-1`.

src/TSMapEditor/Constants.cs
```diff
@@ -91,7 +91,7 @@ public static class Constants
         public static int MaxHouseTechLevel = 10;
 
         public const int MAX_MAP_LENGTH_IN_DIMENSION = 512;
-        public const int NO_OVERLAY = 255; // 0xFF
+        public const int NO_OVERLAY = -1;
```
Rampastring, line 94: Since 255 still appears in parts of the code in a fitting context for `NO_OVERLAY`, instead of changing this I think it is better to have two constants. Something along the lines of `NO_OVERLAY_LEGACY` and `NO_OVERLAY_EXTENDED`.

The `continue` case in `ReadOverlays` could be moved inside the `if` clause's two cases for clarity.

Rampastring (changes requested):

src/TSMapEditor/Initialization/MapWriter.cs
ZivDero: Uhh.... no, `char` is definitely 1 byte (a `short` is 2, and an `int`/`long` is 4).
A `char` is also *signed* by default (to make an unsigned `char`, you type `unsigned char`).
`255` as a literal is an `int`, which when cast to `char` will overflow and end up as `-1`.

src/TSMapEditor/Constants.cs
ZivDero: In fact, 255 appears practically nowhere. Remember that in the game OverlayType is an enum, so it has the size of an int, and `OVERLAY_NONE` is set to equal `-1`. `255` only ever appears because we save `-1` as a `char`, where `-1` is in fact `0xFF`.

Rampastring (approved):

PR #245, opened 2025-06-14
Add base nodes and terrain objects to info display

JoyfulShush:
Adds the ability to show base nodes and terrain objects in the info display when hovering over a cell.

Base nodes:
image

Terrain Objects:
image

### DIFF
--- a/src/TSMapEditor/Models/MapTile.cs
+++ b/src/TSMapEditor/Models/MapTile.cs
@@ -5,6 +5,7 @@ using TSMapEditor.GameMath;
 using TSMapEditor.Models.Enums;
 using TSMapEditor.Models.MapFormat;
 using TSMapEditor.Rendering;
+using TSMapEditor.UI;
 
 namespace TSMapEditor.Models
 {
@@ -254,7 +255,7 @@ namespace TSMapEditor.Models
             {
                 action(waypoint);
             }
-        }       
+        }
 
         public SubCell GetSubCellClosestToPosition(Point2D position, bool onlyOccupiedCells)
         {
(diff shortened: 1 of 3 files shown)


src/TSMapEditor/Models/MapTile.cs
```diff
@@ -447,6 +447,40 @@ public void ChangeTileIndex(int newTileIndex, byte newSubTileIndex)
             SubTileIndex = newSubTileIndex;
         }
 
+        public BaseNode GetBaseNode(Map map)
```
Rampastring: `MapTile` should not have a dependency on `Map`. Introducing one makes the `MapTile - Map` relation a circular dependency, which should be avoided unless there's a good reason not to.

An easy solution to fix this is moving the method to the `Map` class instead of having it in `MapTile` - it doesn't even use any data of `MapTile`, just properties of the full `Map` instance.

When implemented in `Map`, you can also consider iterating on `GraphicalBaseNodes` instead of raw house base nodes, which simplifies the implementation and saves you from having to iterate the `BuildingTypes` list on each iteration, since `GraphicalBaseNode` saves the building type of the base node.

Rampastring (changes requested):

src/TSMapEditor/Models/MapTile.cs
JoyfulShush: Had no idea about `GraphicalBaseNodes`, that helps a lot, thanks for the excellent tip!

* Moved from MapTile to Map, now it only needs a cell coords to know what cell we're asking base nodes for
* Improved the function to use `GraphicalBaseNodes` list
* Improved the tile info display to include the color of the house.

src/TSMapEditor/Models/MapTile.cs
```diff
@@ -445,7 +445,7 @@ public void ChangeTileIndex(int newTileIndex, byte newSubTileIndex)
             TileImage = null;
             TileIndex = newTileIndex;
             SubTileIndex = newSubTileIndex;
-        }
+        }
```
Rampastring: Please revert this change.

Rampastring (changes requested):
Looks good aside from it adding some whitespace to MapTile.cs.

src/TSMapEditor/UI/TileInfoDisplay.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using Microsoft.Xna.Framework;
 using Rampastring.XNAUI;
 using Rampastring.XNAUI.XNAControls;
+using SharpDX.XAPO.Fx;
```
Rampastring: Looks like UnIntelliSense slipped in an extra using clause.

Rampastring (changes requested):

Rampastring (approved):

PR #257, opened 2025-08-09
Add the ability to create an alliance between multiple houses

JoyfulShush:
Introduces the ability to create an alliance between multiple houses quickly via a new button in the Houses window.

Additionally, refactors how houses store alliances; instead of doing so as a list of strings, houses now take a list of Houses that they are allied with. Also handles loading and saving house alliances from or into maps.

Lastly, when a house is deleted through the Houses Window, all other houses remove that house from their allies list if they had it.

### DIFF
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -1180,6 +1180,26 @@ namespace TSMapEditor.Initialization
                     houseType.XNAColor = house.XNAColor;
                 }
             }
+            
+            foreach (var house in map.Houses)
+            {
+                var houseSection = mapIni.GetSection(house.ININame);
+                if (houseSection != null)
+                {
+                    var allies = houseSection.GetListValue("Allies", ',', s => s);
+                    foreach (var ally in allies)
+                    {
+                        var alliedHouse = map.Houses.Find(house => house.ININame == ally);
+                        if (alliedHouse == null)
+                        {
+                            AddMapLoadError($"House with name {ally} was not found when loading up allies for house {house.ININame}. Skipping the house from being loaded.");
+                            continue;
+                        }
+
+                        house.Allies.Add(alliedHouse);
+                    }
+                }
+            }
 
             Logger.Log("Houses read successfully.");
         }
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -1181,6 +1181,7 @@ namespace TSMapEditor.Initialization
                 }
             }
             
+            // Once all houses were loaded into WAE, read and set up the alliances of each house as a list of allied houses
             foreach (var house in map.Houses)
             {
                 var houseSection = mapIni.GetSection(house.ININame);
@@ -1192,7 +1193,7 @@ namespace TSMapEditor.Initialization
                         var alliedHouse = map.Houses.Find(house => house.ININame == ally);
                         if (alliedHouse == null)
                         {
-                            AddMapLoadError($"House with name {ally} was not found when loading up allies for house {house.ININame}. Skipping the house from being loaded.");
+                            AddMapLoadError($"House with name {ally} was not found when loading up allies for the house {house.ININame}. Skipping the house from being loaded.");
                             continue;
                         }
 
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -1187,13 +1187,13 @@ namespace TSMapEditor.Initialization
                 var houseSection = mapIni.GetSection(house.ININame);
                 if (houseSection != null)
                 {
-                    var allies = houseSection.GetListValue("Allies", ',', s => s);
-                    foreach (var ally in allies)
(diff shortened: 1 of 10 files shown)


src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -1180,6 +1180,27 @@ public static void ReadHouses(IMap map, IniFile mapIni)
...
+            foreach (var house in map.Houses)
+            {
+                var houseSection = mapIni.GetSection(house.ININame);
+                if (houseSection != null)
+                {
+                    var allies = houseSection.GetListValue("Allies", ',', s => s);
+                    foreach (var ally in allies)
+                    {
+                        var alliedHouse = map.Houses.Find(house => house.ININame == ally);
+                        if (alliedHouse == null)
```
ZivDero: Please use explicit types for basic types.

src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -1180,6 +1180,27 @@ public static void ReadHouses(IMap map, IniFile mapIni)
...
+                        {
+                            AddMapLoadError($"House with name {ally} was not found when loading up allies for the house {house.ININame}. Skipping the house from being loaded.");
+                            continue;
+                        }
```
ZivDero: This could maybe become a function and be used as a predicate? Not necessarily.

src/TSMapEditor/UI/Windows/ConfigureAlliesWindow.cs
```diff
@@ -39,8 +39,15 @@ public override void Initialize()
...
         {
-            string allies = string.Join(',', new string[] { house.ININame }.Concat(checkBoxes.FindAll(chk => chk.Checked).Select(chk => chk.Text)));
-            house.Allies = allies;
+            List<House> alliedHousesList = [house];
+            var alliedHouses = checkBoxes.FindAll(chk => chk.Checked).Select(chk => (House)chk.Tag);
+
+            foreach (var alliedHouse in alliedHouses)
+            {
+                alliedHousesList.Add(alliedHouse);
+            }
```
ZivDero: A single-line foreach is better off without braces.

src/TSMapEditor/UI/Windows/GenerateStandardHousesWindow.cs
```diff
@@ -50,7 +50,7 @@ private void AddHousesFromEditorRulesSectionAndHide(string sectionName)
                 var house = houses[i];
                 house.ID = i;
                 house.Edge = "North";
-                house.Allies = house.ININame;
+                house.Allies = [house];
```
ZivDero, line 53: This looks so bizzare in C#... where are we going with this new instance xD
No need to do anything, just thoughts out loud.

src/TSMapEditor/UI/Windows/HousesWindow.cs
```diff
@@ -183,6 +192,12 @@ private void BtnDeleteHouse_LeftClick(object sender, System.EventArgs e)
                             throw new InvalidOperationException("Failed to delete HouseType associated with house " + editedHouse.ININame);
                     }
 
+                    // Remove this house from all other houses that were allied to it
+                    foreach (var house in map.Houses)
+                    {
+                        house.Allies.Remove(editedHouse);
+                    }
```
ZivDero: Same for braces.

src/TSMapEditor/UI/Windows/HousesWindow.cs
```diff
@@ -243,6 +258,11 @@ private void BtnMakeHouseNotRepairBuildings_LeftClick(object sender, EventArgs e
             dialog.YesClickedAction = _ => map.Structures.FindAll(s => s.Owner == editedHouse).ForEach(b => b.AIRepairable = false);
         }
 
+        private void BtnSetAlliances_LeftClick(object sender, EventArgs e)
```
ZivDero: => perhaps?

src/TSMapEditor/UI/Windows/SetAlliancesWindow.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                foreach (var otherHouse in alliedHouses)
+                {
+                    if (otherHouse == house)
+                        continue;
+
+                    if (!house.Allies.Contains(otherHouse))
+                    {
+                        house.Allies.Add(otherHouse);
+                    }
```
ZivDero: Same for braces.

ZivDero (changes requested):

src/TSMapEditor/UI/Windows/GenerateStandardHousesWindow.cs
JoyfulShush: Yeah, for whateve reason a house must be allied to itself. It's bizzare that it is not by default and that you have to add this line /shrug

src/TSMapEditor/Initialization/MapLoader.cs
JoyfulShush: Originally what I did was put the logic as a method in the House class. But because `AddMapLoadError` is a private function that belongs to `MapLoader`, if you do that, you cannot use that to add the errors.

Though, on a second thought, I guess I can make it so the function returns a list of issues that I can iterate and add as a map loading error. Would you like me to do so?

PR #268, opened 2025-09-08
Add default config INI file translation support

KuroNoSeiHai:
Add Default/UI/Windows config INI file translation support


Rampastring (approved):
Good work! `translate($$$TextBlockBegin` is not supported, but I'll be implementing something for that.

PR #269, opened 2025-09-09
Translate top bar menu

JoyfulShush:
Translates the entire Top Bar menu, including all missing translated texts. 

Does not apply for inner children, mostly due to most `.ini` files being translated; instead, different PRs will handle all other instances of `.cs` files for windows, cursors/mutations, and remaining texts.

### DIFF
--- a/src/TSMapEditor/UI/TopBar/TopBarMenu.cs
+++ b/src/TSMapEditor/UI/TopBar/TopBarMenu.cs
@@ -93,7 +93,7 @@ namespace TSMapEditor.UI.TopBar
 
             var fileButton = new MenuButton(WindowManager, fileContextMenu);
             fileButton.Name = nameof(fileButton);
-            fileButton.Text = Translate(this, "File.MenuButton", "File");
+            fileButton.Text = Translate(this, "File.Header", "File");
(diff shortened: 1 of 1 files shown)


src/TSMapEditor/UI/TopBar/TopBarMenu.cs
```diff
@@ -93,27 +93,27 @@ public override void Initialize()
 
             var fileButton = new MenuButton(WindowManager, fileContextMenu);
             fileButton.Name = nameof(fileButton);
-            fileButton.Text = "File";
+            fileButton.Text = Translate(this, "File.MenuButton", "File");
```
Rampastring: I think the "menu button" might be unnecessary context here. It could be removed, since it is followed by the sub-items anyway, giving it more context. Alternatively, it could replaced with something like `File.Header`.

```suggestion
            fileButton.Text = Translate(this, "File", "File");
```

Also goes for the rest of the headers / menu buttons.

Rampastring (changes requested):
Looks mostly good, one comment to address.

Rampastring (approved):

PR #280, opened 2025-10-14
Fix Ares Event 87 and 88 repeated

KuroNoSeiHai:
Origin Event 87 and 88 use the same section. Makes it work uncorrectly.


Rampastring (approved):
Good catch. This fix has been included in `yr` branch.

PR #283, opened 2025-10-22
Revise SpriteFont with Chinese ranges, cyrillic rendering, and symbols rendering

SadPencil:
## Update by @ApocBruh 

update adds some commonly used Chinese characters and improves some characters, including: Add characters:
- U+201C -> U+201D, U+2022, U+2026, U+2014, U+2018 -> U+2019: General Punctuation 

Fix some character to look more good:
"#" in light font and bold font :
Old:
Image
image
New:
Image
image

", $, ª, «, ¶, º, » in bold font:
Old:
Image
New:
Image

## Update by @SadPencil 

This PR replaces the main font's Chinese character range with a carefully selected set. The original range was poorly chosen, resulting in both missing necessary characters and the inclusion of overly rare characters.

This update ensures that the limited space is used to provide the most complete and relevant character set possible, without deleting any other non-Chinese characters. **Note that it still contains fewer Chinese characters compared with a sprite font made just for Chinese.**

The new font contains 6942 Chinese characters as follows:

- Chinese characters defined in GB2312-90.

- Chinese characters defined in Tong Yong Gui Fan Han Zi Biao (levels 1 & 2 only) AND (logic AND) in Unicode 1.1.

## Update by @ApocBruh 

Image

Top: before. Bottom: After

See https://github.com/CnCNet/xna-cncnet-client/issues/860 for the issue description


SadPencil:
P.S. The predecessor 48e6c9633abce01944cc8724efac479a62519793 has a known issue https://github.com/CnCNet/xna-cncnet-client/issues/860 and this PR does not fix it

Edit: now fixed

PR #287, opened 2025-11-01
Copy and Paste Triggers, Trigger Events and Trigger Actions

JoyfulShush:
This PR introduces the ability to copy and paste whole triggers, trigger events and trigger actions, preserving most of their values.

The data is serialized and copied into the Clipboard, allowing users to copy them even across multiple instances of WAE, where they are deserialized and recreated there, allowing users to copy across maps.

This also introduces a new helpers file - StreamHelper. This helper streamlines common operations, such as reading and writing bools, strings, and numbers in multiple variations into one place for easy and simple reusability, which serves most of the code in the PR (and to lesser extend - pasting terrain).

Some caveats to note:
* Triggers never copy their linked triggers. Each trigger regenerates a new ID from the map alongside with a new tag to accompany it.
* Trigger events and actions can be copied and pasted together. However, the trigger is standalone, so copying a trigger will override copied events and actions.
* In order to support pasting triggers in a more intuitive way, the triggers' List Box was changed to behave just like events and action List Boxes, where right clicking anywhere in them opens the context menu, not just when hovering over a valid trigger.

I've put a lot of effort into testing it - hopefully I didn't miss anything.

### DIFF
--- a/src/TSMapEditor/StreamHelpers.cs
+++ b/src/TSMapEditor/StreamHelpers.cs
@@ -1,0 +1,124 @@
+﻿
+
+using System;
+using System.IO;
+using System.Reflection.Metadata;
+using System.Text;
+using TSMapEditor.Models;
+
+namespace TSMapEditor
+{
+    public class StreamHelperReadException : Exception
+    {
+        public StreamHelperReadException(string message) : base(message)
+        {
+        }
+    }    
+
+    public static class StreamHelpers
+    {
+        public static int ReadInt(Stream stream)
+        {
+            byte[] buffer = new byte[8];
+
+            if (stream.Read(buffer, 0, 4) != 4)
+                throw new StreamHelperReadException("Failed to read integer from stream: end of stream");
+
+            return BitConverter.ToInt32(buffer, 0);
+        }
+
+        public static long ReadLong(Stream stream)
+        {
+            byte[] buffer = new byte[8];
+
+            if (stream.Read(buffer, 0, 8) != 8)
+                throw new StreamHelperReadException("Failed to read integer from stream: end of stream");
+
+            return BitConverter.ToInt64(buffer, 0);
+        }
+
+        public static string ReadASCIIString(Stream stream)
+        {
+            int length = ReadInt(stream);
+            byte[] stringBuffer = new byte[length];
+            if (stream.Read(stringBuffer, 0, length) != length)
+                throw new StreamHelperReadException("Failed to read string from stream: end of stream");
+
+            if (length == -1)
+                return null;
+
+            string result = Encoding.ASCII.GetString(stringBuffer);
+            return result;
+        }
+
+        public static byte[] ASCIIStringToBytes(string str)
+        {
+            byte[] buffer = new byte[sizeof(int) + str.Length];
+            Array.Copy(BitConverter.GetBytes(str.Length), buffer, sizeof(int));
+            byte[] stringBytes = Encoding.ASCII.GetBytes(str);
+            Array.Copy(stringBytes, 0, buffer, sizeof(int), stringBytes.Length);
+            return buffer;
+        }
+
+        public static string ReadUnicodeString(Stream stream)
+        {
+            int length = ReadInt(stream);
+
+            if (length == -1)
+                return null;
+
+            byte[] stringBuffer = new byte[length];
+            if (stream.Read(stringBuffer, 0, length) != length)
+                throw new StreamHelperReadException("Failed to read Unicode string from stream: end of stream");            
+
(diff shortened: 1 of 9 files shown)


src/TSMapEditor/StreamHelpers.cs
```diff
@@ -0,0 +1,120 @@
...
+    public static class StreamHelpers
+    {
+        public static int ReadInt(Stream stream)
+        {
+            byte[] buffer = new byte[8];
+
+            if (stream.Read(buffer, 0, 4) != 4)
+                throw new StreamHelperReadException("Failed to read integer from stream: end of stream");
+
+            return BitConverter.ToInt32(buffer, 0);
```
Rampastring: `new` allocates memory on the heap, which is expensive. Avoid repeated `new` calls when possible by using `Span`s and `stackalloc` instead.

`CopiedMapEntry` previously avoided the issue by reusing one array instance that was only ever allocated once, but due to this class being static, that is not possible here.

```suggestion
            Span<byte> buffer = stackalloc byte[4];

            if (stream.Read(buffer) != sizeof(int))
                throw new StreamHelperReadException("Failed to read integer from stream: end of stream");

            return BitConverter.ToInt32(buffer);
```

Adjust the other functions accordingly.

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -2471,4 +2550,156 @@ private void Map_TriggersChanged(object sender, EventArgs e)
             }
         }
     }
+
+    public static class CopiedTriggerData
```
Rampastring: 1) `TriggersWindow.cs` is already huge. Split this into its own file, maybe under the `TSMapEditor/Models` folder.

2) For flexibility, make this class and all its members non-static, and have TriggersWindow create and contain an instance of this class.

src/TSMapEditor/UI/Windows/TriggersWindow.cs
```diff
@@ -2471,4 +2550,156 @@ private void Map_TriggersChanged(object sender, EventArgs e)
...
+                memoryStream.Position += stringLength;
+            }
+        }
+
+        public static bool HasTriggerActionDataInClipboard()
+        {
+            return HasTriggerActionOrEventData(true);
+        }
+
+        public static bool HasTriggerEventDataInClipBoard()
```
Rampastring: For consistency with others.

```suggestion
        public static bool HasTriggerEventDataInClipboard()
```

Rampastring (changes requested):
Looks mostly good, but needs some work.

src/TSMapEditor/StreamHelpers.cs
JoyfulShush: TIL. Had no idea that's a thing. Thanks!

src/TSMapEditor/StreamHelpers.cs
```diff
@@ -0,0 +1,121 @@
...
+                throw new StreamHelperReadException("Failed to read integer from stream: end of stream");
+
+            return BitConverter.ToInt32(buffer);
+        }
+
+        public static long ReadLong(Stream stream)
+        {
+            Span<byte> buffer = stackalloc byte[8];
+
+            if (stream.Read(buffer) != sizeof(int))
```
Rampastring: Looks like a copy/paste mistake that currently makes this fail.

```suggestion
            if (stream.Read(buffer) != sizeof(long))
```

src/TSMapEditor/StreamHelpers.cs
```diff
@@ -0,0 +1,121 @@
...
+                throw new StreamHelperReadException("Failed to read integer from stream: end of stream");
```
Rampastring:
```suggestion
                throw new StreamHelperReadException("Failed to read long integer from stream: end of stream");
```

src/TSMapEditor/StreamHelpers.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            return BitConverter.ToInt64(buffer);
+        }
+
+        public static string ReadASCIIString(Stream stream)
+        {
+            int length = ReadInt(stream);
+            Span<byte> stringBuffer = stackalloc byte[length];
+            if (stream.Read(stringBuffer) != length)
+                throw new StreamHelperReadException("Failed to read string from stream: end of stream");
```
Rampastring:
```suggestion
                throw new StreamHelperReadException("Failed to read ASCII string from stream: end of stream");
```

Rampastring (changes requested):

Rampastring (approved):

PR #288, opened 2025-11-05
Add the ability to place waypoints immediately on position when holding Shift

JoyfulShush:
When holding shift, skips the dialog window and instead immediately places the next free waypoint with the ~~default color (none, which default to purple)~~ last picked color.

Respects used, available and max waypoints and shows the appropriate messages using the same flow as with the existing flow.

### DIFF
--- a/src/TSMapEditor/UI/CursorActions/PlaceWaypointCursorAction.cs
+++ b/src/TSMapEditor/UI/CursorActions/PlaceWaypointCursorAction.cs
@@ -19,7 +19,16 @@ namespace TSMapEditor.UI.CursorActions
 
         public override void LeftClick(Point2D cellCoords)
         {
-            PlaceWaypointWindow.Open(cellCoords);
+            bool isShiftHeldDown = Keyboard.IsShiftHeldDown();
+            if (isShiftHeldDown)
+            {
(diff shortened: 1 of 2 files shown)


Rampastring:
Would it be better to use the last-used color instead of using the default color?

JoyfulShush:
Turns out that it already does that. Makes sense, since the window is only initialized once and it remembers the value of the color that was last picked by the user. I'll update the description to match.

src/TSMapEditor/UI/CursorActions/PlaceWaypointCursorAction.cs
```diff
@@ -19,7 +19,16 @@ public PlaceWaypointCursorAction(ICursorActionTarget cursorActionTarget) : base(
 
         public override void LeftClick(Point2D cellCoords)
         {
-            PlaceWaypointWindow.Open(cellCoords);
+            bool isShiftHeldDown = Keyboard.IsShiftHeldDown();
+            if (isShiftHeldDown)
```
Rampastring:
```suggestion
            if (Keyboard.IsShiftHeldDown())
```

Rampastring (changes requested):
Looks good aside from having an useless state variable in `PlaceWaypointCursorAction`.

Rampastring (approved):

PR #291, opened 2025-11-08
Add support for disabling local AI Triggers

JoyfulShush:
Introduces the ability to enable or disable local AI Triggers from the AI Triggers Window. Reads and writes to the map file by utilizing the `AITriggerTypesEnable` INI section.

### DIFF
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -1099,6 +1099,23 @@ namespace TSMapEditor.Initialization
             }
 
             Logger.Log("AITriggerTypes read successfully.");
+
+            section = mapIni.GetSection("AITriggerTypesEnable");
+            if (section == null)
+                return;
+
+            foreach (var kvp in section.Keys)
+            {
+                string aiTriggerININame = kvp.Key;
+
+                var aiTrigger = map.AITriggerTypes.Find(aiTriggerType => aiTriggerType.ININame == aiTriggerININame);
+                if (aiTrigger == null)
+                    continue;
+
+                aiTrigger.Disabled = !Conversions.BooleanFromString(kvp.Value, true); // enabled by default
+            }
+
+            Logger.Log("AITriggerTypesEnable read successfully.");
         }
 
         public static void ReadHouseTypes(IMap map, IniFile mapIni)
--- a/src/TSMapEditor/Initialization/MapLoader.cs
+++ b/src/TSMapEditor/Initialization/MapLoader.cs
@@ -1112,7 +1112,7 @@ namespace TSMapEditor.Initialization
                 if (aiTrigger == null)
                     continue;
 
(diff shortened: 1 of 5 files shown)


src/TSMapEditor/Models/AITriggerType.cs
```diff
@@ -91,6 +91,7 @@ public AITriggerType(string iniName)
         public bool Easy { get; set; } = true;
         public bool Medium { get; set; } = true;
         public bool Hard { get; set; } = true;
+        public bool Disabled { get; set; } = false;
```
Rampastring: Looking through the overall code, you can get somewhat simpler logic by reversing this, in other words changing this to `public bool Enabled { get; set; } = true;`.

That way you can avoid some negations in the code, and it would match with the `AITriggerTypesEnable` section.

The check box in `AITriggersWindow` should then also be called `chkEnabled`.

Rampastring (changes requested):

src/TSMapEditor/Models/AITriggerType.cs
JoyfulShush: It's what I originally had, but thought about Triggers having "Disabled" and thought you'd want to keep consistency.

However I do agree it's easier to navigate through the code without it.

Flipped the logic and did some basic testing, looks like it's working well.

src/TSMapEditor/Models/AITriggerType.cs
```diff
@@ -91,6 +91,7 @@ public AITriggerType(string iniName)
         public bool Easy { get; set; } = true;
         public bool Medium { get; set; } = true;
         public bool Hard { get; set; } = true;
+        public bool Enabled { get; set; } = false;
```
Rampastring: We surely want AI triggers to be enabled by default.

```suggestion
        public bool Enabled { get; set; } = true;
```

src/TSMapEditor/Initialization/MapLoader.cs
```diff
@@ -1099,6 +1099,23 @@ public static void ReadAITriggerTypes(IMap map, IniFile mapIni)
...
+
+            foreach (var kvp in section.Keys)
+            {
+                string aiTriggerININame = kvp.Key;
+
+                var aiTrigger = map.AITriggerTypes.Find(aiTriggerType => aiTriggerType.ININame == aiTriggerININame);
+                if (aiTrigger == null)
+                    continue;
+
+                aiTrigger.Enabled = Conversions.BooleanFromString(kvp.Value, true);
```
Rampastring: It's usually best practice to default values to existing values when they are read from INI, to avoid "resetting key" bugs in case the code is made to read multiple INI files at a later date.

```suggestion
                aiTrigger.Enabled = Conversions.BooleanFromString(kvp.Value, aiTrigger.Enabled);
```

Rampastring (changes requested):

Rampastring (approved):

PR #295, opened 2025-11-15
Add support for Team Delays

JoyfulShush:
Allows users to manage their map's TeamDelays key through WAE, which is something done somewhat common to control the production speed of the AI in many missions.

Introduces a new script. Running it opens the difficulty settings management window. For now, this only allows the user to instruct WAE to add a key for TeamDelays to the map, in which case they will have to provide values for each difficulty level:
image

Once applied, if the user saves, WAE will add `[General]` section if not exists yet, and will add the `TeamDelays=` key to it with the appropriate values. Likewise, if the user instructs WAE to not override team delays, WAE will remove `TeamDelays=` key from it, and also the `[General]` section if there are no other keys in it.

Supports loading and saving the values from and to any map being worked on.


Rampastring (changes requested):
I'm unsure whether this functionality actually belongs to WAE as difficulty settings are purely text-based and have no graphical impact. A script that adds all the difficulty code to the map for easy editing might already cut it together with the existing `File -> Open With Text Editor`.

That being said, if we go forward with this, the implementation needs some work.

Model-wise, `TeamDelays` does not belong into the Map, but to Rules. `TeamDelays=` exists in `Rules.ini` after all. It should also be of type `int[]` or `List<int>`, because `TeamDelays=` is a list of integers.

Then, modifying `TeamDelays` would change it in the Rules data instead of modifying the map model.

To save the modified Rules data, WAE would need a way to write specific rules overrides into the map when the map is saved. I don't think a function for this exists currently because WAE has no existing interface where you'd modify the values of Rules.ini keys.

Rampastring (approved):

PR #297, opened 2025-11-19
Add Phobos Features

CnCRAZER:
I think this is everything.. Not 100% positive though.

Used FA2SP 1.6.3 `FAData.ini` as reference as well as the docs for phobos v0.4.


Metadorius:
I mean could still commit those? just that the main source should be Phobos docs

Metadorius:
@Rampastring @ZivDero @JoyfulShush What do you prefer, Phobos compat. amendments being only in Phobos docs or here too (but possibly not the most recent)?

ZivDero:
> @Rampastring @ZivDero @JoyfulShush What do you prefer, Phobos compat. amendments being only in Phobos docs or here too (but possibly not the most recent)?

I think they should be here. If they're sometimes slightly outdated that's fine

CnCRAZER:
> I mean could still commit those? just that the main source should be Phobos docs

I don't mind committing to both but I think it's best to have them confirmed and looked over before committing to WAE. Can open back up if desired

JoyfulShush:
I agree, this is what we did with Vinifera as well. It's not like they get outdated often.

PR #299, opened 2025-11-22
Add verbose script action texts to script actions list

JoyfulShush:
Adds handling of most script action types that are handled via the script parameter value, showing the appropriate values on the script actions list.

Scripts actions will now show dynamic text, as following:
* Script actions with a numeric value (waypoint number, cells, etc.) retain their number. e.g. `(10)`.
* Script actions with preset options now show the selected preset option text. e.g. `(Keep Transports, Lose Units)`
* Script actions with a certain handled type, such as local variables, houses, units/structures, animations etc. will now show the appropriate text based on the type. e.g. `(GDI)`, `(Game started)`, `(Power Plant - Least threat)`, etc.
* Script actions that accept no values (typically shows "Use 0" on the parameter) will now simply show the script action's name, without any additional information. For example: `#3 - Deploy`.

See below example of script actions in action:
image

Additionally, removes unhandled, unused types for `TeamType` for `Change Team` and `ScriptType` for `Change Script`. They are both unhandled correctly by WAE, and it serves no real purpose; if it will ever be implemented correctly, can be expanded to support the types.

### DIFF
--- a/src/TSMapEditor/UI/Windows/ScriptsWindow.cs
+++ b/src/TSMapEditor/UI/Windows/ScriptsWindow.cs
@@ -865,7 +865,7 @@ namespace TSMapEditor.UI.Windows
             string actionEntryText;
             string textDetails = null;
 
-                ScriptAction action = GetScriptAction(entry.Action);
+            ScriptAction action = GetScriptAction(entry.Action);
             if (action == null)
             {                
                 actionEntryText = $"#{index} - Unknown";
@@ -891,7 +891,7 @@ namespace TSMapEditor.UI.Windows
                     case TriggerParamType.LocalVariable:
                         var localVar = map.LocalVariables.GetElementIfInRange(entry.Argument);
                         if (localVar != null)
-                            textDetails = $"({localVar.Name})";
+                            textDetails = $"({localVar.Index} - {localVar.Name})";
                         break;
 
                     case TriggerParamType.HouseType:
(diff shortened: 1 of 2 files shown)


Rampastring (commented):
I think it'd be useful to still show the number even when showing descriptions. For example, with locals, it's useful to see which number it is, so instead of

`Set Local (Waves remaining)`

You'd have

`Set Local (1 Waves remaining)`

or

`Set Local (1 - Waves remaining)`.

JoyfulShush:
Excellent idea. Done. I went with the `Set Local (1 - Waves remaining)` variant.

Rampastring:
I meant doing that for all parameter types, not just locals.

JoyfulShush:
Added for all operations that have reference to an indexed entity - building types, animations, presets, and house types.

src/TSMapEditor/UI/Windows/ScriptsWindow.cs
```diff
@@ -944,11 +946,72 @@ private void EditScript(Script script)
...
+                        break;
+
+                    case TriggerParamType.Unknown:
+                        // special handling: script actions that have no type but still have presets should be handled by showing the preset value
+                        // otherwise we'll show no text after the name of the action
+                        if (hasValidPresetOption) 
+                            goto default;
+
+                        textDetails = null;
+                        break;
```
Rampastring: What is the point of this case? Can't it be entirely handled by the `default` case?

Rampastring (approved):

PR #301, opened 2025-11-23
Add script action count warning and allow loading all script actions

JoyfulShush:
The game does not support over 50 script actions in one script. In order to raise awareness and to make the logic consistent with other WAE components, such as triggers, this PR applies the following changes:
* WAE now loads all script actions for each script in the map file, rather than stopping at 50.
* WAE now warns the user whenever a script is saved with over 50 script actions.

image

### DIFF
--- a/src/TSMapEditor/Models/Map.cs
+++ b/src/TSMapEditor/Models/Map.cs
@@ -2092,6 +2092,18 @@ namespace TSMapEditor.Models
                         Constants.TS_WAYPT_SPECIAL));
             }
 
+            // Check for scripts that have more than 50 Script Actions. This is unsupported by the game.
+            const int maxScriptActionCount = 50;
+            foreach (var script in Scripts)
+            {
+                if (script.Actions.Count > maxScriptActionCount)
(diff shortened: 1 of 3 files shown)


src/TSMapEditor/Models/Map.cs
```diff
@@ -2092,6 +2092,18 @@ public List<string> CheckForIssues()
                         Constants.TS_WAYPT_SPECIAL));
             }
 
+            // Check for scripts that have more than 50 Script Actions. This is unsupported by the game.
+            const int maxScriptActionCount = 50;
+            foreach (var script in Scripts)
+            {
+                if (script.Actions.Count > maxScriptActionCount)
+                {
+                    issueList.Add(string.Format(Translate(this, "CheckForIssues.ScriptTooManyActions",
```
Rampastring, line 2101: Please update `Translation_en.ini` with the new string.

Rampastring (changes requested):

Rampastring (approved):

PR #302, opened 2025-12-27
Default House AI Repair Logic

JoyfulShush:
Adds a new key to houses used by WAE, which denotes whether future buildings they place have the "AI Repairable" flag automatically set or not. 

This key is set to `true` or `false` when mappers use the "Enable/Disable Building Repair" buttons, respectively. New text is also added to show it, both in the confirmation message and also in the house statistics.

This is useful since in most cases, mappers that use the Enable Building Repair typically want to make sure all structures for that house (typically used by AI) are automatically repaired, but currently it only applies to current structures, not future structures.

Automatically saves and loads as a key from/to the map for each house via INI reading/writing logic.

### DIFF
--- a/src/TSMapEditor/Config/Translations/en/Translation_en.ini
+++ b/src/TSMapEditor/Config/Translations/en/Translation_en.ini
@@ -738,6 +738,7 @@ HousesWindow.EnableAIRepairs.Title=Are you sure?
 HousesWindow.EnableAIRepairs.Description.v2=This enables the "AI Repairable" flag on all buildings of the house, which makes the AI repair them.@@Additionally, this will cause the "AI Repairable" flag to be automatically enabled for all buildings you place for this house.@@No un-do is available. Do you wish to continue?
 HousesWindow.DisableAIRepairs.Title=Are you sure?
 HousesWindow.DisableAIRepairs.Description.v2=This disables the "AI Repairable" flag on all buildings of the house, which makes the AI NOT repair them.@@Additionally, this will cause the "AI Repairable" flag to be automatically disabled for all buildings you place for this house.@@No un-do is available. Do you wish to continue?
+
 InfantryOptionsWindow.lblHeader.Text=Infantry Options
 InfantryOptionsWindow.lblSelectedInfantry.Text=Selected Infantry:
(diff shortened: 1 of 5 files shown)


Rampastring (changes requested):
This PR modifies translated strings, but does not update the reference translation.

Furthermore, from the contribution guideline:

> If your code significantly modifies existing strings so that their meaning is changed (fixing typos etc. does not count), you need to update the translation key for that string to signal the change to translators. A good way to do this is to append an incrementing version identifier to the translation key, or increment the key if it already exists. For example, if you modified the `ExpandMapWindow.InvalidWidth.Title` string in a significant way, you'd change its translation key to `ExpandMapWindow.InvalidWidth.Title.v2`. If then, you returned and changed its meaning again at a later date, you'd change the translation key to `ExpandMapWindow.InvalidWidth.Title.v3`.

This PR significantly modifies `EnableAIRepairs.Description` and `DisableAIRepairs.Description`.

JoyfulShush:
Updated translations per contribution guidelines.

src/TSMapEditor/Config/Translations/en/Translation_en.ini
```diff
@@ -736,8 +736,10 @@ HousesWindow.NoHouseSelected.Title=No House Selected
 HousesWindow.NoHouseSelected.Description=Select a house first.
 HousesWindow.EnableAIRepairs.Title=Are you sure?
 HousesWindow.EnableAIRepairs.Description=This enables the "AI Repairs" flag on all buildings of the house, which makes the AI repair them.@@No un-do is available. Do you wish to continue?
```
Rampastring: You should also remove the old translation strings that aren't used anymore (`HousesWindow.EnableAIRepairs.Description` and `HousesWindow.DisableAIRepairs.Description`).

Rampastring (changes requested):

src/TSMapEditor/Config/Translations/en/Translation_en.ini
JoyfulShush: Removed the old translations.

Rampastring (approved):

PR #303, opened 2026-02-06
Add warning for triggers using Building Exists incorrectly

JoyfulShush:
Adds a new warning when saving maps that should help mappers in case they use the Building Exists event incorrectly. Only applies to SP maps.

When saving the map, if the map is likely to be a SP map, checks all triggers and sees if they have the Building Exists event. Any trigger that have this event is checked for the following:
* Gets the trigger's house. If the house is under the player control, skips checking the trigger.
* Checks what is requested in the Building Exists. Then makes sure the AI should have it at some point via one of the following checks:
  * Building for the house is pre-placed in the map.
  * House has the building as a Base Node. (note: we don't check that that those base nodes will actually be activated or can be built - too complex)
  * House has a unit that deploys into that building (e.g. MCV -> ConYard) in one of its TeamTypes.

Added English translations keys.

### DIFF
--- a/src/TSMapEditor/Misc/MapIssueChecker.cs
+++ b/src/TSMapEditor/Misc/MapIssueChecker.cs
@@ -1,6 +1,7 @@
 ﻿using Rampastring.Tools;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using TSMapEditor.CCEngine;
 using TSMapEditor.Models;
 using TSMapEditor.Models.Enums;
@@ -12,6 +13,7 @@ namespace TSMapEditor.Misc
     {
         private const int EnableTriggerActionIndex = 53;
         private const int DisableTriggerActionIndex = 54;
+        private const int BuildingExistsTriggerActionIndex = 32;
         private const int TriggerParamIndex = 1;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -12,6 +13,7 @@ public static class MapIssueChecker
     {
         private const int EnableTriggerActionIndex = 53;
         private const int DisableTriggerActionIndex = 54;
+        private const int BuildingExistsTriggerActionIndex = 32;
```
Rampastring: Building Exists is not an action, but an event.

```suggestion
        private const int BuildingExistsTriggerEventIndex = 32;
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
                             trigger.Name, trigger.HouseType));
             }
 
+            // Check for triggers in SP that have Building Exists event but house belongs to an AI
+            // that doesn't ever build or deploys a unit into the building
```
Rampastring:
```suggestion
            // that doesn't ever build or deploy a unit into the building
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
...
+                    // Get the house of the trigger
+                    var house = map.Houses.Find(house => house.ININame == trigger.HouseType);
+                    
+                    if (house == null)
+                        continue; // will be reported by the check for invalid owners
+
+                    foreach (var condition in trigger.Conditions)
+                    {
+                        // Check if current condition is Building Exists condition
+                        if (condition.ConditionIndex != BuildingExistsTriggerActionIndex) continue;
```
Rampastring: We don't really do this "in-line" style in WAE.

Fix the others accordingly.

```suggestion
                        if (condition.ConditionIndex != BuildingExistsTriggerActionIndex)
                            continue;
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
...
+
+                        // This check only checks AI houses as they're controlled by mappers in SP missions
+                        if (house.PlayerControl) continue;
```
Rampastring: Checking this for each condition is pointless, because the `house` is the same for each condition. Move this check to around line 404, outside of the `foreach` loop.

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
...
+                        // Get the condition's building index
+                        int buildingIndex = int.Parse(condition.Parameters[1]);
+
+                        // Try to find a pre-placed structure on the map with the index belonging to the trigger's house
+                        // If found - all good, house has this structure and trigger is valid
+                        var building = map.Structures.Find(structure => 
+                            structure.ObjectType.Index == buildingIndex &&
+                            structure.Owner == house);
+                        
+                        if (building != null) continue;
```
Rampastring: The variable `building` seems to be unused after this check, so we can simplify the code here a bit.

```suggestion
                        // Try to find a pre-placed structure on the map with the index belonging to the trigger's house
                        // If found - all good, house has this structure and trigger is valid
                        if (map.Structures.Exists(structure => structure.ObjectType.Index == buildingIndex && structure.Owner == house))
                            continue;
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
...
+                        
+                        // Get the building type (rules configuration)
+                        var buildingType = map.Rules.BuildingTypes.Find(buildingType => buildingType.Index == buildingIndex);
+                        if (buildingType == null)
+                        {
+                            issueList.Add(string.Format(Translate(map, "CheckForIssues.BuildingExists.InvalidBuildingType",
+                                "Could not find buildingType with index {0} as was specified in trigger '{1}'"),
+                            buildingIndex, trigger.Name));
+                            continue;
+                        }
```
Rampastring: The building index is literally an index into the type list, there is no need for a string-based lookup.

```suggestion
                        if (buildingIndex < 0 || buildingIndex >= map.Rules.BuildingTypes.Count)
                        {
                            issueList.Add(string.Format(Translate(map, "CheckForIssues.BuildingExists.InvalidBuildingType",
                                "No building type index {0} exists as was specified in trigger '{1}'"),
                            buildingIndex, trigger.Name));
                            continue;
                        }
                        
                        var buildingType = map.Rules.BuildingTypes[buildingIndex];
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
...
+
+                        // check the house's base nodes to see if the AI is intended to build it at some point
+                        // If found - all good, house will build this when it can (not checking activation or MCV etc.)
+                        var baseNode = house.BaseNodes.Find(baseNode => baseNode.StructureTypeName == buildingType.ININame);
+                        if (baseNode != null) continue;
```
Rampastring: Can be simplified, like the existing building check.

```suggestion
                        // Check the house's base nodes to see if the AI is intended to build it at some point
                        // If found - all good, house will build this when it can (we assume they have a ConYard etc.)
                        if (house.BaseNodes.Exists(baseNode => baseNode.StructureTypeName == buildingType.ININame))
                            continue;
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
...
+
+                        // Check if there is some techno that can deploy into this structure
+                        // If so, check if it's pre-placed or is in a task force in the TeamType used by the house
+                        bool hasDeployedTechno = false;
+                        var unitsThatDeployToBuilding = map.Rules.UnitTypes
+                            .FindAll(unitType => unitType.DeploysInto == buildingType.ININame)
+                            .Select(unitType => unitType.ININame)
+                            .ToList();
```
Rampastring, line 442: Some of the comparisons below can be simplified if we make this a list of unit types rather than strings.

```suggestion
                        var unitsThatDeployToBuilding = map.Rules.UnitTypes
                            .FindAll(unitType => unitType.DeploysInto == buildingType.ININame)
                            .ToList();
```

Rampastring (changes requested):

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,104 @@ public static List<string> CheckForIssues(Map map)
...
+                                            continue;
+
+                                        if (unitsThatDeployToBuilding.Contains(technoTypeEntry.TechnoType.ININame))
+                                            hasDeployedTechno = true;
+                                    }
+                                }
+                            }
+                        }
+
+                        if (hasDeployedTechno) continue;
```
Rampastring: We can do some functional programming here too to simplify the code flow and avoid loops within loops. This also allows us to get rid of the `hasDeployedTechno` variable.


```suggestion
                        if (unitsThatDeployToBuilding.Count > 0)
                        {
                            if (map.Units.Exists(u => u.Owner == house && unitsThatDeployToBuilding.Contains(u.UnitType)))
                                continue;

                            if (map.TeamTypes.Exists(teamType => 
                            {
                                if (teamType.HouseType != house.HouseType)
                                    return false;

                                if (teamType.TaskForce == null)
                                    return false;

                                return Array.Exists(teamType.TaskForce.TechnoTypes, tte => tte != null && tte.TechnoType.WhatAmI() == RTTIType.UnitType && unitsThatDeployToBuilding.Contains((UnitType)tte.TechnoType));
                            }))
                            {
                                continue;
                            }
                        }
```

Rampastring (changes requested):

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,91 @@ public static List<string> CheckForIssues(Map map)
...
+                        
+                        // Get the condition's building index
+                        int buildingIndex = int.Parse(condition.Parameters[1]);
+
+                        // Try to find a pre-placed structure on the map with the index belonging to the trigger's house
+                        // If found - all good, house has this structure and trigger is valid
+                        if (map.Structures.Exists(structure => 
+                            structure.ObjectType.Index == buildingIndex &&
+                            structure.Owner == house))
+                            continue;
```
Rampastring: From the contribution guideline (`Contributing.md`):

> Braceless code block bodies should be made only when both code block head and body are single line.

There are two ways to fix this. This is my preferred way (we don't live in the 90s with 800x600 resolution anymore, there's no problem having lines be a bit longer):

```suggestion
                        if (map.Structures.Exists(structure => structure.ObjectType.Index == buildingIndex && structure.Owner == house))
                            continue;
```

This is the alternative:

```suggestion
                        if (map.Structures.Exists(structure => 
                            structure.ObjectType.Index == buildingIndex &&
                            structure.Owner == house))
                        {
                            continue;
                        }
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,91 @@ public static List<string> CheckForIssues(Map map)
...
+                        
+                        if (buildingIndex < 0 || buildingIndex >= map.Rules.BuildingTypes.Count)
+                        {
+                            issueList.Add(string.Format(Translate(map, "CheckForIssues.BuildingExists.InvalidBuildingType",
+                                "No building type index {0} exists as was specified in trigger '{1}'"),
+                            buildingIndex, trigger.Name));
```
Rampastring: Incorrect indentation.

```suggestion
                                buildingIndex, trigger.Name));
```

src/TSMapEditor/Misc/MapIssueChecker.cs
```diff
@@ -388,6 +390,91 @@ public static List<string> CheckForIssues(Map map)
...
+
+                                if (teamType.TaskForce == null)
+                                    return false;
+
+                                return Array.Exists(teamType.TaskForce.TechnoTypes, tte =>
+                                    tte != null &&
+                                    tte.TechnoType.WhatAmI() == RTTIType.UnitType &&
+                                    unitsThatDeployToBuilding.Contains((UnitType)tte.TechnoType));
+                            }))
+                                continue;
```
Rampastring, line 463: From the contribution guideline (`Contributing.md`):

> Braceless code block bodies should be made only when both code block head and body are single line.

```suggestion
                            }))
                            {
                                continue;
                            }
```

Rampastring (changes requested):

Rampastring (approved):

PR #307, opened 2026-03-24
Add zh-Hans translations

KuroNoSeiHai:
Add zh-Hans translations for WAE


Rampastring (approved):
I updated the comment in `Translation_zh-Hans.ini` because the English reference translation file's comments weren't applicable.

I don't understand Chinese, but it looks like great work!

PR #309, opened 2026-04-10
Add clone handling for clone-able entities

JoyfulShush:
Introduce `Helpers.HandleCloneNaming` to centralize clone naming behavior: if a name contains a number, the first numeric part is incremented. Otherwise a localized `" (Clone)"` suffix is appended.

Update AITriggerType, Script, TaskForce, TeamType, and Trigger to use the new helper instead of hardcoded string for each entity.

Add/update the `CloneName` translation key in English and zh-Hans (consolidating previous per-type entries) to support localization.

### DIFF
--- a/src/TSMapEditor/Helpers.cs
+++ b/src/TSMapEditor/Helpers.cs
@@ -855,7 +855,7 @@ namespace TSMapEditor
         /// Otherwise, a localized " (Clone)" suffix is appended to the name.
         /// If there are multiple numbers in the name, only the first is incremented.
         /// </summary>
-        public static string HandleCloneNaming(string name)
+        public static string GetNameForClone(string name)
         {
             string[] parts = name.Split(" ");
             int numPart = -1;            
@@ -864,7 +864,7 @@ namespace TSMapEditor
             {                
                 if (int.TryParse(parts[i], out numPart) && numPart >= 0)
                 {
-                    parts[i] = (numPart + 1).ToString();
+                    parts[i] = (numPart + 1).ToString(CultureInfo.InvariantCulture);
                     return string.Join(" ", parts);
                 }
             }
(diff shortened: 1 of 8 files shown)


src/TSMapEditor/Helpers.cs
```diff
@@ -847,5 +847,29 @@ public static string DifficultyToTranslatedString(Difficulty difficulty)
...
+        public static string HandleCloneNaming(string name)
+        {
+            string[] parts = name.Split(" ");
+            int numPart = -1;            
+
+            for (int i = 0; i < parts.Length; i++)
+            {                
+                if (int.TryParse(parts[i], out numPart) && numPart >= 0)
+                {
+                    parts[i] = (numPart + 1).ToString();
```
Rampastring: Best to use `InvariantCulture` to make sure that system locale does not affect the conversion.

This can require adding `System.Globalization` to the `using`s at the top of the file.

```suggestion
                    parts[i] = (numPart + 1).ToString(CultureInfo.InvariantCulture);
```

src/TSMapEditor/Helpers.cs
```diff
@@ -847,5 +847,29 @@ public static string DifficultyToTranslatedString(Difficulty difficulty)
...
         }
+
+        /// <summary>
+        /// Shared helper function used by Triggers, TaskForces, Scripts, TeamTypes, and AI Triggers.
+        /// Used to generate the name of the instance during the cloning process of those entities.
+        /// If the provided name contains a number, the number is incremented and the name is returned as is. 
+        /// Otherwise, a localized " (Clone)" suffix is appended to the name.
+        /// If there are multiple numbers in the name, only the first is incremented.
+        /// </summary>
+        public static string HandleCloneNaming(string name)
```
Rampastring: I'd prefer `GetNameForClone` for the name of this function. Update the refererences accordingly.

Rampastring (changes requested):
Looks good, needs a couple small changes.

Rampastring (approved):

PR #322, opened 2026-07-09
Add headless, scriptable Random Map Generation (RMG) CLI mode

BlackgamerzVN:
## ⚠️ NOT BUILD-VERIFIED — READ BEFORE MERGING ⚠️

This PR was developed in a sandboxed environment with **no network access to nuget.org**, and the sandbox's global NuGet config also has a stale, unrelated local source (`C:\INI Parser`) that breaks `dotnet restore`/`dotnet build` before the real dependencies can even be resolved. This was confirmed independently in two separate sandboxes (this session and the collaborating client-side session), and the `C:\INI Parser` path has zero references anywhere in this repo (grepped) — it's leftover machine cruft, not a project dependency.

**As a result, none of the code in this PR has been compiled or run.** All of it was written and cross-checked manually against existing call sites, signatures, and types in the repo (`MapSetup.cs`, `TerrainGeneratorConfigWindow.cs`, `TerrainGenerationMutation.cs`, `Map.cs`, `MapWriter.cs`, `Theater.cs`, etc.), but that is not a substitute for a real build + smoke test.

**Before merging, someone with a working build environment should:**
1. `dotnet build` the solution and fix any compile errors.
2. Run a real smoke test, e.g.:
   ```
   WorldAlteringEditor.exe --generate-map --game-directory "<path to a TS/RA2/YR install>" --theater Temperate --width 100 --height 100 --players 4 --seed 12345 --output C:\temp\test.map
   ```
   and confirm it exits 0, writes a valid `.map` file, and that the file loads correctly back in the normal WAE GUI (terrain looks sane, player waypoints exist, resource overlays are present and don't overlap water/roads).
3. Sanity-check both `--symmetry rotational` (default) and `--symmetry mirror` for a few player counts (2, 3, 4).
4. Sanity-check `--algorithm wfc` (see below) against the default `--algorithm scatter`, ideally on a theater with a terrain generator preset that has 2+ tile groups (so WFC actually has multiple domains to work with).

### Risk / verification status by piece, so reviewers know what's what
| Piece | Status |
|---|---|
| Headless CLI plumbing (`--generate-map`, hidden window, arg parsing) | Core to this PR's purpose; same age/maturity as everything else here — **unverified**, but a straightforward, contained change (`Program.cs` / `GameClass.cs`). |
| `--algorithm scatter` (default) | Directly reuses the existing, already-shipped `TerrainGenerationMutation` engine exactly as the GUI's Terrain Generator tool calls it, just over the whole map instead of a brush selection. Lowest risk of the terrain paths — closest to "known-good code, new caller." |
| Symmetry (`rotational`/`mirror`) + wedge-balanced resources | New geometry/math logic, moderate risk, but self-contained (no changes to existing mutation/rendering code paths). |
| `--algorithm wfc` (opt-in, non-default) | **Newest and highest-risk piece.** A brand-new from-scratch algorithm (`WaveFunctionCollapseSolver.cs`) with no prior art in this codebase to cross-check signatures against — only manually reasoned through, not just manually diffed against existing patterns. Strongly recommend reviewing/testing this path last and separately from the rest, and it's safe to ship the rest without it if it needs more iteration (it's fully opt-in; default behavior is untouched). |

## What this adds

A new `--generate-map` headless CLI mode so an external process (e.g. a CnCNet game lobby client) can invoke WAE non-interactively to produce a randomized, playable `.map` file — no visible window, structured stdout/exit-code contract for automation.

### Why a hidden window instead of a pure console app
`TheaterGraphics` allocates real `Texture2D` GPU objects in its constructor, so loading theater tile/overlay/smudge data requires a live MonoGame `GraphicsDevice` — this is true even outside of rendering. Decoupling that is a much bigger refactor with no real benefit for this use case, so this PR keeps a real MonoGame window but never shows/activates it, and skips `MainMenu`/all interactive UI when `--generate-map` is passed.

### What it generates
- A blank map of the requested theater/size (`Map.InitNew`, same path as the GUI's "new map" flow).
- Terrain variety via one of two selectable algorithms (`--algorithm`, see below and the risk table above), both built on the existing `TerrainGenerationMutation` engine (previously only ever driven by a user-brushed selection) through a new lightweight non-UI `IMutationTarget` stub (`HeadlessMutationTarget`).
- Symmetric player start `Waypoint`s (0..N-1), with a choice of:
  - **Rotational** (default): N-fold rotational symmetry around the map center.
  - **Mirror**: left/right axis-mirror symmetry across a vertical line through the map center (players paired up; an odd leftover player is placed on the axis).
- Resource fields (Tiberium/ore/gem overlays), placed with per-player-wedge balance: patch centers are sampled only within one symmetric "fundamental domain" and then replicated via the same rotation/reflection transform used for player starts, with patch count scaled by playable map area — instead of pure independent random scatter per cell.
- Deterministic output given the same `--seed` (fixed a pre-existing bug where `TerrainGenerationMutation` always reseeded internally from `DateTime.Now`, which would have silently broken determinism for the new `--seed` flag; now takes an optional explicit seed, backward-compatible for existing GUI callers).

### Terrain algorithm: `--algorithm scatter|wfc` (default `scatter`)
- **`scatter`** (default, unchanged from earlier iterations of this PR): runs `TerrainGenerationMutation` once over the whole map, with all configured tile groups competing independently per cell.
- **`wfc`** (new, opt-in, highest-risk piece — see table above): a real Wave Function Collapse solver (`WaveFunctionCollapseSolver.cs`) — a from-scratch implementation of the "simple tiled model" variant of the algorithm popularized by [mxgmn/WaveFunctionCollapse](https://github.com/mxgmn/WaveFunctionCollapse) (MIT licensed; no code from that repo is reused here, only the published algorithm description: observe the lowest-Shannon-entropy cell, collapse it to a weighted-random single state, propagate the resulting adjacency constraints via BFS, detect contradictions). It divides the map into 6x6-cell blocks and solves for which single `TerrainGeneratorTileGroup` governs each block, using the theater's **real** `LATGrounds` ground/base adjacency data (plus tileset index 0, the theater's default base/clear tileset) to build the compatibility matrix — so neighboring blocks only ever get tile groups the game's own tileset rules say are actually allowed to be adjacent. Each block is then painted with the existing `TerrainGenerationMutation` engine restricted to just that tile group, and the existing auto-LAT blending (already enabled on `HeadlessMutationTarget`) smooths the boundaries between adjacent blocks, the same way manual GUI tile painting does. Falls back to `scatter` automatically if the solver hits a contradiction, if a block's generation throws, or if the active preset has fewer than 2 tile groups.

## Finalized CLI contract

```
WorldAlteringEditor.exe --generate-map
    --game-directory <path>     (required; game install dir containing Rules.ini / mix files)
    --theater <name>            (required; e.g. Temperate, Urban, Snow, Desert, Lunar)
    --width <n>                 (required, > 0)
    --height <n>                (required, > 0)
    --players <n>                (required, 2-8)
    --seed <n>                   (optional; omit for a random seed)
    --gamemode <name>           (optional; accepted/logged only, see below)
    --symmetry <rotational|mirror>  (optional; default "rotational")
    --algorithm <scatter|wfc>   (optional; default "scatter")
    --output <absolute path>   (required; .map file to write)
```

Both space-separated (`--flag value`) and `--flag=value` forms are accepted for every flag, to match what the calling client (`xna-cncnet-client`'s `ExternalProcessRandomMapGenerator`) already expects.

- **Success**: exit code `0`, stdout: `Map generation succeeded. Output written to: <path>`
- **Failure**: exit code `1`, stderr with an error message (missing/invalid args, missing theater preset, I/O failure, etc.)
- The window is never shown/activated in this mode.

## Known open design questions (flagged, not blocking)
- `--gamemode` is currently accepted and logged only — it does not yet select a specific `TerrainGeneratorPresets.ini` section or bias resource density/ratio. Open question for whoever wires up game-mode-specific behavior.
- Ore vs. gem ratio is not controlled — all `OverlayType.Tiberium == true` types (ore and gem pooled together) are picked uniformly at random per patch.
- Mirror-mode resource balance is only left/right (2 replicas), not fully per-player, for player counts > 2 — see inline comment in `PlaceResources` / plan notes for details. Rotational mode doesn't have this limitation.
- Resource density constant (`~900 cells per patch attempt` per symmetric replica) and WFC block size (`6x6 cells`) are first-pass constants, not tuned against real game balance/visual-quality data.

## Files changed
- `src/TSMapEditor/Initialization/HeadlessMapGenerator.cs` (new) — `HeadlessRmgOptions` (CLI parsing/validation), `HeadlessMapGenerator` (orchestration: terrain, resources, player starts, save), `SymmetryMode` and `TerrainAlgorithm` enums.
- `src/TSMapEditor/Initialization/WaveFunctionCollapseSolver.cs` (new) — the generic WFC solver used by the `wfc` terrain algorithm.
- `src/TSMapEditor/Initialization/HeadlessMutationTarget.cs` (new) — minimal non-UI `IMutationTarget` stub.
- `src/TSMapEditor/Program.cs` — parses `--generate-map` and its options.
- `src/TSMapEditor/Rendering/GameClass.cs` — hides the window and runs headless generation instead of showing `MainMenu` when `--generate-map` is passed.
- `src/TSMapEditor/Mutations/Classes/TerrainGenerationMutation.cs` — added optional explicit `seed` parameter (backward compatible) to fix determinism.

## License note
This repo is GPLv3. This PR is a straightforward feature addition to the existing GPLv3 codebase (not a fork), so no additional licensing action is needed beyond normal repo contribution. The new `WaveFunctionCollapseSolver.cs` reimplements a publicly documented algorithm from scratch; it does not copy code from the MIT-licensed reference implementation, and MIT is compatible with GPLv3 in any case.

### DIFF
--- a/src/TSMapEditor/Initialization/HeadlessMapGenerator.cs
+++ b/src/TSMapEditor/Initialization/HeadlessMapGenerator.cs
@@ -27,6 +27,25 @@ namespace TSMapEditor.Initialization
     }
 
     /// <summary>
+    /// Terrain generation strategy used by the headless RMG pipeline.
+    /// </summary>
+    public enum TerrainAlgorithm
+    {
+        /// <summary>The original approach: run <see cref="Mutations.Classes.TerrainGenerationMutation"/>
+        /// once over the whole map with all configured tile groups competing independently per cell.</summary>
+        Scatter,
+
+        /// <summary>
+        /// Divide the map into blocks and use a Wave Function Collapse solver (see
+        /// <see cref="WaveFunctionCollapseSolver"/>) to decide which tile group governs each
+        /// block, respecting the theater's real LAT ground/base adjacency rules so neighboring
+        /// blocks are only assigned tile groups that are actually allowed to sit next to each
+        /// other. Produces more spatially coherent terrain clusters than <see cref="Scatter"/>.
+        /// </summary>
+        WaveFunctionCollapse
+    }
+
+    /// <summary>
     /// Parameters for headless, non-interactive random map generation.
     /// </summary>
     public class HeadlessRmgOptions
@@ -62,6 +81,13 @@ namespace TSMapEditor.Initialization
         /// Defaults to <see cref="SymmetryMode.Rotational"/>.
         /// </summary>
         public SymmetryMode Symmetry { get; set; } = SymmetryMode.Rotational;
+
+        /// <summary>
+        /// Terrain generation strategy. Defaults to <see cref="TerrainAlgorithm.Scatter"/>
+        /// (the original, already-shipped behavior) for backwards compatibility; pass
+        /// "--algorithm wfc" to opt into the newer Wave-Function-Collapse-based layout.
+        /// </summary>
+        public TerrainAlgorithm Algorithm { get; set; } = TerrainAlgorithm.Scatter;
 
         /// <summary>Path (including file name) that the generated .map file should be written to.</summary>
         public string OutputPath { get; set; }
@@ -132,6 +158,13 @@ namespace TSMapEditor.Initialization
                             options.Symmetry = SymmetryMode.Mirror;
                         else if (value.Equals("rotational", StringComparison.OrdinalIgnoreCase))
                             options.Symmetry = SymmetryMode.Rotational;
+                        break;
+                    case "algorithm":
+                        if (value.Equals("wfc", StringComparison.OrdinalIgnoreCase) ||
+                            value.Equals("wavefunctioncollapse", StringComparison.OrdinalIgnoreCase))
+                            options.Algorithm = TerrainAlgorithm.WaveFunctionCollapse;
+                        else if (value.Equals("scatter", StringComparison.OrdinalIgnoreCase))
+                            options.Algorithm = TerrainAlgorithm.Scatter;
                         break;
                     case "output":
                         options.OutputPath = value;
@@ -226,14 +259,14 @@ namespace TSMapEditor.Initialization
 
                 var mutationTarget = new HeadlessMutationTarget(map);
 
-                GenerateTerrain(map, mutationTarget, theater, random);
+                GenerateTerrain(map, mutationTarget, theater, random, options.Algorithm);
                 PlaceResources(map, random, options.PlayerCount, options.Symmetry);
                 PlacePlayerStarts(map, options.PlayerCount, options.Symmetry);
 
                 map.AutoSave(options.OutputPath);
 
                 Logger.Log($"Headless RMG: generated {options.Width}x{options.Height} {options.Theater} map " +
-                    $"with {options.PlayerCount} player starts ({options.Symmetry} symmetry, seed {seed}, gamemode {options.GameMode ?? "(none)"}) -> {options.OutputPath}");
+                    $"with {options.PlayerCount} player starts ({options.Symmetry} symmetry, {options.Algorithm} terrain algorithm, seed {seed}, gamemode {options.GameMode ?? "(none)"}) -> {options.OutputPath}");
 
                 return null;
             }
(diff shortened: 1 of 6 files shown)


ZivDero:
That looks... quite plain. As far as the maps it can make go

Metadorius:
bru

BlackgamerzVN:
fuckkk I forgot to fork it

Metadorius:
image
that looks like a fork?

ZivDero:
> image that looks like a fork?

I feel like something went wrong in launching AI to work on it -> the "could not build" thing. That would indicate the PR was made automatically, too, though xD

BlackgamerzVN:
It is, and AI fucker sent it directly to HERE xD

ZivDero:
> It is, and AI fucker sent it directly to HERE xD

You should probably reconsider giving it such broad permissions

BlackgamerzVN:
noted.

BlackgamerzVN:
MOTHERFUCKER

BlackgamerzVN:
Myeah expect future slops

Rampastring:
> TheaterGraphics allocates real Texture2D GPU objects in its constructor, so loading theater tile/overlay/smudge data requires a live MonoGame GraphicsDevice — this is true even outside of rendering. Decoupling that is a much bigger refactor with no real benefit for this use case, so this PR keeps a real MonoGame window but never shows/activates it, and skips MainMenu/all interactive UI when --generate-map is passed.

Despite the mountain of slop that this PR is, the above reminded me to decouple logical tile and overlay reading from MonoGame, so maps and tile data can be loaded without allocating GPU textures. It had been on my mind for a while.

That refactor has now been done, so adding any "CLI" style functionality should be easier and more efficient in the future.

PR #324, opened 2026-07-10
Run cursor action pre-draw logic before capturing the object under th…

NayiemW:
Since the `MapUI` / `MapView` split, `TechnoUnderCursor` has been passed into `MapView.Draw` before the active cursor action’s `PreMapDraw` logic runs.

This breaks placement range indicators because building placement previews assign `TechnoUnderCursor` inside `PreMapDraw`. Since the renderer receives the previous value, it does not see the newly assigned object during that frame. The property is then reset during the next update cycle, so range indicators never appear while placing buildings such as pillboxes.

Previously, the relevant order was effectively:

```text
PreMapDraw
→ read TechnoUnderCursor
→ draw map and range indicators
→ PostMapDraw
```

After the split, it became:

```text
read and pass TechnoUnderCursor
→ PreMapDraw
→ draw map
→ PostMapDraw
```

This change moves the `PreMapDraw` invocation into `MapUI.Draw`, before `TechnoUnderCursor` is passed to the renderer. `MapView.Draw` no longer invokes it a second time.

The corrected order is therefore:

```text
PreMapDraw
→ pass the updated TechnoUnderCursor to MapView
→ draw map and range indicators
→ PostMapDraw
```

`PostMapDraw` remains in its existing location, so the intended call order around map rendering is preserved while ensuring placement previews can display their range indicators.

### DIFF
--- a/src/TSMapEditor/Rendering/MapView.cs
+++ b/src/TSMapEditor/Rendering/MapView.cs
@@ -1614,10 +1614,9 @@ namespace TSMapEditor.Rendering
 
         public void Draw(bool isActive, TechnoBase technoUnderCursor, MapTile tileUnderCursor, CursorAction cursorAction)
         {
-            if (isActive && tileUnderCursor != null && cursorAction != null)
-            {
-                cursorAction.PreMapDraw(tileUnderCursor.CoordsToPoint());
-            }
+            // Note: the cursor action's PreMapDraw is invoked by MapUI.Draw before this method,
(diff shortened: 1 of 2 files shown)


Rampastring:
Good catch. Due to conflicts I couldn't merge this PR directly, but I fixed the bug in https://github.com/CnCNet/WorldAlteringEditor/commit/5c5b5f348bbb3656257444755e374d18baced447 and added you as a co-author to the commit, while also leaving out the AI's excessive comments in `MapView.cs`.

PR #330, opened 2026-08-11
Increase MAX_MAP_LENGTH_IN_DIMENSION to 2048

SethGekco:
Anarchy


Rampastring:
Tiberian Sun and Yuri's Revenge limit this to 512.

If a major engine extension lifts the limit in the future, increasing the limit can be considered, but even then it needs to be configurable to avoid breaking compatibility with vanilla.

Also, WAE's current renderer does not work for maps that exceed 16384 pixels in one dimension.

